PR #3, opened 2026-07-09
authz: nested docsets override ancestor grants (read/write carve-out)

aakarim:
Fixes the over-permissive read behavior flagged after v0.2.0: an identity with a read/write grant on the root docset (`openlore` = `/`, granted `ro` to everyone incl. anonymous) could read/write **every** nested docset (`/alfie`, `/jared`, …).

## Change
A docset is now an access boundary — the **most-specific docset** covering a path governs access. A grant on an ancestor docset no longer reaches into a nested docset the identity has no grant on.

- `mostSpecificDocset()`; `grantForPath` resolves via it and requires a grant on the governing docset (write carve-out).
- `scopedReadFS` carries all docset roots as boundaries; `within()` denies a path when a more-specific ungranted docset covers it; `ReadDir` always filters children (hides carved-out nested docsets).
- passkey `/lore` browser now reads through the same per-identity scoped session FS (no more raw-FS + flat prefix allow-list bypass).
- `validateGrants` rejects two docsets sharing a display root (ambiguous boundary).

## Verification
- `go build`, `go vet`, `go test ./...` green; added carve-out + duplicate-root regression tests.
- Deployed to prod and verified live: anonymous and non-granted agents can no longer read `/alfie`, `/jared`, `/knowledge`; explicit grants still work.

> Note: v0.2.0 is already published from `main` without this fix. Suggest cutting **v0.2.1** (or moving the release) after merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,25 +201,25 @@ OpenLore is built on [Wish](https://github.com/charmbracelet/wish) from Charmbra
 
 | Command | Description |
 |---------|-------------|
-| `whoami` | Print your identity and lore |
+| `whoami` | Print your identity |
 | `lore` | Introspection dispatcher (run `lore` for subcommands) |
-| `lore docsets` | List the docsets you can access, their paths, and attributes |
+| `lore docsets` | List the docsets you can access, their grant, paths, and attributes |
 
 `lore docsets` prints an aligned, greppable table:
 
 ```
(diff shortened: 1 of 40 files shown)


aakarim:
Superseded by a clean PR based on main (this branch conflicted because main independently integrated the grant model).

PR #10, opened 2026-07-13
fix: mount production docset containers

aakarim:
## Summary

Prepare openlore.sh for the production docset layout migration by replacing the old per-docset mounts with organizational container mounts:

- `/agent` → `published/agent`
- `/user` → `published/user`
- `/channel` → `published/channel`

`lore.json` continues to own child docsets and grants, so direct writes to the container roots remain denied.

This is required for canonical paths such as `/agent/jared` and `/channel/general`; the old embedded mounts would conflict with the compatibility aliases `/jared` and `/knowledge`.

## Validation

- workflow-equivalent linux/amd64 build with this config embedded
- `go test ./...`
- production dry-run identified the old mount collision before any data was changed

### DIFF
--- a/openlore.sh/openlore.yml
+++ b/openlore.sh/openlore.yml
@@ -57,23 +57,12 @@ tokens:
   refresh_ttl: 720h
 
 folders:
-  - name: knowledge
-    path: ./published/knowledge
-  - name: jared
-    path: ./published/jared
-  - name: claw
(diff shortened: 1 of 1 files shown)


aakarim:
Closing in favor of a single writable-root mount design. Production namespace topology should not be duplicated in the binary's embedded config and lore.json.

PR #35, opened 2026-08-19
Add delegated identity permissions dashboard

aakarim:
## Summary
- add a passkey-authenticated self-service dashboard at `/settings/permissions`
- let principals update delegate docset/capability denials and disconnect delegates with session-bound CSRF protection
- revoke delegated refresh-token chains on disconnect and audit updates/removals
- show effective access and recorded CIMD/JWKS/mTLS verification levels
- share one embedded light/dark stylesheet across dashboard, OAuth consent, and passkey pages

## Verification
- `go test ./...`
- `go test -race ./internal/passkeys ./pkg/openlore`
- `go vet ./...`
- `go build ./cmd/openlore`
- browser preview checked at desktop width in light color scheme

### DIFF
--- a/pkg/openlore/settings_permissions.go
+++ b/pkg/openlore/settings_permissions.go
@@ -26,6 +26,7 @@ type permissionToggle struct {
 type delegateView struct {
 	Identity              string
 	DisplayName           string
+	ResolutionError       string
 	ClientAuth            ClientAuthLevel
 	ClientAuthLabel       string
 	ClientAuthVerified    bool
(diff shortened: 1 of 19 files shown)


aakarim (commented):
Reviewed against the [permissions-dashboard plan](/channel/project-openlore/plans/delegated-identity-permissions-dashboard.md) plus the two parent plans. **The plan is implemented in full** — routes (`GET /settings/permissions`, PRG POSTs for update/remove), passkey-session auth with session-bound CSRF, subtractive deny toggles reusing the consent semantics, disconnect with `RevokeDelegation` and revoked-chain counts in the `delegate.remove` audit detail, effective-access preview, client-auth badge (with the new persisted `DelegateEntry.ClientAuth` and monotonic upgrade at token issuance — a sensible gap-fill the plan implied), shared `internal/webstyle` stylesheet applied to dashboard/consent/login/register with light-scheme support, file browser untouched, and all out-of-scope items correctly absent. Build, vet, and the full test suite pass locally.

Nothing unimplemented. Four nonblocking quality findings inline.

pkg/openlore/settings_permissions.go
```diff
@@ -0,0 +1,310 @@
...
+	docsets, capabilities, err := s.permissionsFor(session.Identity, "")
+	if err != nil {
+		http.Error(w, "failed to resolve permissions", http.StatusInternalServerError)
+		return
+	}
+	views := make([]delegateView, 0, len(principal.Delegates))
+	for _, delegate := range principal.Delegates {
+		effectiveDocsets, effectiveCapabilities, err := s.permissionsFor(session.Identity, delegate.Identity)
+		if err != nil {
+			continue
```
aakarim: **Nonblocking:** a delegate whose policy fails to resolve (e.g. it references a role that has since been deleted) is silently skipped, so it vanishes from the dashboard — the one page a user would visit to remove a broken delegate. Consider rendering the row in an error state with the Disconnect button still available.

pkg/openlore/settings_permissions.go
```diff
@@ -0,0 +1,310 @@
...
+			ClientAuthLabel:    clientAuthLabel(ClientAuthLevel(delegate.ClientAuth)),
+			ClientAuthVerified: ClientAuthLevel(delegate.ClientAuth).Verified(),
+			EffectiveDocsets:   effectiveDocsets, EffectiveCapabilities: effectiveCapabilities,
+			DeniedDocsets: difference(docsets, allowedDocsets), DeniedCapabilities: difference(capabilities, allowedCapabilities),
+		}
+		if identity, found := s.findAuthIdentity(delegate.Identity); found {
+			if identity.ClientIDMetadata != nil {
+				view.DisplayName = identity.ClientIDMetadata.PinnedName
+			} else {
+				view.DisplayName = strings.TrimPrefix(identity.Comment, "Display: ")
```
aakarim, line 103: **Nonblocking:** `TrimPrefix` returns the input unchanged when the prefix is absent, so a manually configured delegate identity with e.g. `Comment: "cron agent for reports"` shows that whole comment as its display name. Guard on `strings.HasPrefix` (or `CreatedBy == "oauth"`) and leave DisplayName empty otherwise.

pkg/openlore/settings_permissions.go
```diff
@@ -0,0 +1,310 @@
...
+		return fmt.Errorf("principal %q not found", session.Identity)
+	}, func() (map[string]any, error) {
+		if s.refreshStore == nil {
+			return map[string]any{"revoked_refresh_chains": 0}, nil
+		}
+		revoked, err := s.refreshStore.RevokeDelegation(session.Identity, actor)
+		return map[string]any{"revoked_refresh_chains": revoked}, err
+	})
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
```
aakarim, line 188: **Nonblocking:** if the `after` callback (refresh-chain revocation) fails, `updateAuth` has already persisted the delegate removal and swapped it live, but the user gets a 400 with a raw error and no redirect — it looks like the disconnect failed when it actually succeeded (minus revocation, which is already recorded in the audit detail). Suggest logging the revocation error and redirecting anyway, or at least distinguishing this from a genuine 400 (it's a 5xx-shaped failure).

pkg/openlore/tokenendpoint.go
```diff
@@ -206,6 +211,12 @@ func (t *tokenEndpoint) handleAuthorizationCode(w http.ResponseWriter, r *http.R
 			return
 		}
 	}
+	if c.Actor != "" && t.delegateAuth != nil {
+		if err := t.delegateAuth.recordDelegateClientAuth(c.Subject, c.Actor, clientAuth); err != nil {
```
aakarim, line 217: **Nonblocking:** a failure to record the client-auth badge (a cosmetic upgrade write to `lore.json`) turns into a 500 on the token endpoint and blocks the user's login. Best-effort seems safer: log and continue issuing the token. Separately, `recordDelegateClientAuth` emits the same `delegate.update` audit event type as user-initiated permission edits — audit consumers can't tell a badge upgrade from a permissions change; consider `delegate.client_auth` or a detail field.

pkg/openlore/settings_permissions.go
aakarim: Fixed in cac5182. Post-persist callback failures are audited and logged but no longer reported as if the configuration mutation failed, so disconnect follows PRG after a revocation-store error. Added a regression test for the persisted-removal/error-detail behavior.

pkg/openlore/settings_permissions.go
aakarim: Fixed in cac5182. Delegates whose effective policy cannot be resolved now render in an error state without editable toggles, while retaining the Disconnect action. Added a regression test.

pkg/openlore/tokenendpoint.go
aakarim: Fixed in cac5182. Client-auth badge persistence is now best-effort: failures are logged and token issuance continues. Badge upgrades now emit `delegate.client_auth`, distinct from user permission edits, with regressions for both behaviors.

pkg/openlore/settings_permissions.go
aakarim: Fixed in cac5182. Comment-based display names are now used only when the `Display: ` prefix is present; arbitrary manual comments remain hidden. Covered by the broken-delegate dashboard test.

PR #66, opened 2026-09-03
Folder rules system: Phase 1

aakarim:
## Summary

- add the layered rules registry, compiler, matcher, unification, admission, and validation engine
- add fixed-max size rules, the estimate/v1 tokenizer, and OKF/link adapters with file vs bundle scope
- support top-level and per-docset rules in lore.json, including legacy OKF desugaring and conflict detection
- wire rules into write admission and lore validate while retaining existing OKF metadata behavior
- add openlore.yml rules.growth defaults and reserve unsupported tokenizer configuration

Phase 2 discoverability, Phase 3 .lore/config.yaml files, and Phase 4 persistent initial-size baselines remain out of scope.

## Behavior change

Legacy `okf.enforce: false` now makes the complete OKF rule family (`okf`, `okf/bundle`, and `link/resolves`) report warnings rather than errors. This intentionally preserves a single enforcement switch for the family.

## Verification

- `nix --extra-experimental-features 'nix-command flakes' develop .#default -c sh -c 'go test ./... && go vet ./... && go build ./...'`

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -1414,7 +1414,7 @@ func desugarOKFRules(auth *AuthConfig) error {
 			docset.Rules = map[string]rules.RuleSpec{}
 		}
 		if explicit, ok := docset.Rules["okf"]; ok {
-			if !sameLegacyOKFRule(explicit, expected) {
+			if !explicit.Equal(expected) {
 				return fmt.Errorf("docset %q has conflicting okf and rules.okf configuration", name)
 			}
 		} else {
@@ -1433,26 +1433,4 @@ func desugarOKFRules(auth *AuthConfig) error {
 		auth.Docsets[name] = docset
 	}
 	return nil
-}
-
-func sameLegacyOKFRule(a, b rules.RuleSpec) bool {
-	if a.Use != b.Use || a.Default != b.Default || a.IsEnforcing() != b.IsEnforcing() || len(a.Match) != len(b.Match) || len(a.Exclude) != len(b.Exclude) || len(a.With) != len(b.With) {
-		return false
-	}
-	for i := range a.Match {
-		if a.Match[i] != b.Match[i] {
-			return false
-		}
-	}
-	for i := range a.Exclude {
-		if a.Exclude[i] != b.Exclude[i] {
-			return false
-		}
-	}
-	for key, value := range a.With {
-		if fmt.Sprint(value) != fmt.Sprint(b.With[key]) {
-			return false
-		}
-	}
-	return true
 }
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/aakarim/go-openlore/pkg/rules"
 	"github.com/aakarim/go-openlore/pkg/vfs"
 	"golang.org/x/crypto/ssh"
 	"gopkg.in/yaml.v3"
@@ -62,6 +63,7 @@ type Config struct {
 	// event-bus `hooks` path with middleware on the read/write chains.
 	Shellexec ShellexecConfig
 	Logger    *slog.Logger
+	Rules     RulesConfig
 
 	// Readonly is the global write lock. Default true: the substrate is a
 	// read-only filesystem and no write verbs are available. Set false to
@@ -99,6 +101,10 @@ type Config struct {
 	configFileLoaded   bool
 	embeddedConfigUsed bool
 	warnings           []string
+}
+
+type RulesConfig struct {
+	Growth float64
 }
 
 // Warnings returns non-fatal problems encountered while loading configuration.
@@ -257,11 +263,12 @@ type FilesConfig struct {
 
 // AuthConfig is loaded from lore.json.
 type AuthConfig struct {
-	AllowKeyless    *bool                 `json:"allow_keyless,omitempty"`
-	UnknownIdentity string                `json:"unknown_identity,omitempty"`
-	DefaultCwd      string                `json:"default_cwd,omitempty"`
-	Docsets         map[string]DocsetSpec `json:"docsets"`
-	Roles           map[string]RoleSpec   `json:"roles,omitempty"`
+	AllowKeyless    *bool                     `json:"allow_keyless,omitempty"`
+	UnknownIdentity string                    `json:"unknown_identity,omitempty"`
+	DefaultCwd      string                    `json:"default_cwd,omitempty"`
+	Rules           map[string]rules.RuleSpec `json:"rules,omitempty"`
+	Docsets         map[string]DocsetSpec     `json:"docsets"`
+	Roles           map[string]RoleSpec       `json:"roles,omitempty"`
 	// Default is a legacy authority field retained only for JSON parsing. It is ignored.
 	Default    map[string]string `json:"default,omitempty"`
 	Identities []AuthIdentity    `json:"identities"`
@@ -312,8 +319,9 @@ type JWKSSpec struct {
 
 // DocsetSpec defines a named set of path mappings.
 type DocsetSpec struct {
-	Paths  []PathMapping `json:"paths"`
-	Access DocsetAccess  `json:"access,omitempty"`
+	Paths  []PathMapping             `json:"paths"`
+	Access DocsetAccess              `json:"access,omitempty"`
+	Rules  map[string]rules.RuleSpec `json:"rules,omitempty"`
 	// AgentSkills is ignored. Collections are selected dynamically by xattr.
 	AgentSkills bool `json:"-"`
 	// Aliases are alternate display roots for the first path. They expose the
@@ -503,12 +511,18 @@ type fileConfig struct {
 	Readonly            *bool                  `yaml:"readonly"`
 	WriteConflictPolicy string                 `yaml:"write_conflict_policy"`
 	MaxJobs             int                    `yaml:"max_jobs"`
+	Rules               rulesYAML              `yaml:"rules"`
 	// Tokens + OIDCIssuers are server infrastructure (bearer-token issuance for
 	// the MCP + HTTP API), hence configured here rather than in lore.json.
 	Tokens      *AuthTokensConfig `yaml:"tokens"`
 	OIDCIssuers []OIDCIssuer      `yaml:"oidc_issuers"`
 	Inbox       *inboxYAML        `yaml:"inbox"`
 	Plugins     pluginsYAML       `yaml:"plugins"`
+}
+
+type rulesYAML struct {
+	Growth    *float64 `yaml:"growth"`
+	Tokenizer string   `yaml:"tokenizer"`
 }
 
 type authInfrastructureYAML struct {
@@ -600,6 +614,7 @@ func New(opts ...Option) (Config, error) {
 		Readonly:            true,                           // safe default: read-only substrate
 		WriteConflictPolicy: vfs.DefaultWriteConflictPolicy, // "hash": overwrites are compare-and-swap
 		MaxJobs:             8,                              // bound concurrent async spawn jobs
+		Rules:               RulesConfig{Growth: 1.25},
 		Plugins:             PluginsConfig{Skills: SkillsPluginConfig{RemoteCheckTTL: 60 * time.Second, RemoteTimeout: 3 * time.Second, RemoteMaxBytes: 10 * 1024 * 1024}},
 		Passkeys: PasskeysConfig{
 			Enabled:      true,
@@ -661,6 +676,15 @@ func WithConfigFile(path string) Option {
 		cfg.warnings = append(cfg.warnings, warnings...)
 
 		cfg.configFileLoaded = true
+		if fc.Rules.Tokenizer != "" {
+			return errors.New("rules.tokenizer is not supported yet")
+		}
+		if fc.Rules.Growth != nil {
+			if *fc.Rules.Growth < 1 {
+				return errors.New("rules.growth must be at least 1")
+			}
+			cfg.Rules.Growth = *fc.Rules.Growth
+		}
 
 		if fc.ConfigVersion != "" {
 			cfg.ConfigVersion = fc.ConfigVersion
@@ -1216,6 +1240,9 @@ func LoadAuthConfig(path string) (*AuthConfig, error) {
 // ValidateAuthConfig validates a parsed static authorization policy. Legacy
 // authority fields are deliberately ignored.
 func ValidateAuthConfig(auth *AuthConfig) error {
+	if err := desugarOKFRules(auth); err != nil {
+		return err
+	}
 	if _, ok := auth.Roles["guest"]; ok {
 		return fmt.Errorf("role %q is reserved", "guest")
 	}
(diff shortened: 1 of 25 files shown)


pkg/rules/engine.go
```diff
@@ -0,0 +1,263 @@
...
+
+func (e *Engine) validateFile(ctx context.Context, fsys vfs.FileSystem, bundleRoot, target string, content []byte, bundle *validation.Bundle) []validation.Diagnostic {
+	rules, err := e.Effective(ctx, target)
+	if err != nil {
+		return []validation.Diagnostic{{Path: relative(bundleRoot, target), Line: 1, Column: 1, Severity: validation.SeverityError, Rule: "rules/config", Message: err.Error()}}
+	}
+	var diagnostics []validation.Diagnostic
+	for _, rule := range rules {
+		if rule.Member.Manifest().Scope == ScopeBundle {
+			if bundle == nil || len(bundle.Files) == 0 || target != bundle.Files[0].AbsolutePath || !bundleMatches(rule, bundle) {
```
aakarim: **Blocking (1): bundle rules anchored on `Files[0]`.**

Bundle-scope rules only fire when `target == bundle.Files[0].AbsolutePath`, and the effective rule set is the one resolved for that first file. `Files[0]` is whatever `WalkDir` yields first in lexical order — not the directory the caller asked to validate. Two failure modes:

- `lore validate /` with `okf` docsets at `/docs` and `/wiki` and a root `README.md`: `Files[0]` is `/README.md`, outside every docset, so no `okf/bundle`/`link/resolves` runs anywhere. Reproduced: `/` → `0 errors, 0 warnings` while `/docs` and `/wiki` each report an `openlore/broken-link` error.
- Phase 3: if `Files[0]` sits in a subfolder with its own `.lore/config.yaml`, the bundle rules for `/docs` are the ones effective at `/docs/aaa/`.

Plan §3.4 (updated): resolve `Effective(bundle.Root)` once, evaluate each `ScopeBundle` rule exactly once over the whole bundle, and drop the `Files[0]` check from the members (`okfrule/okf.go:43`, `link/link.go:43`). When the root is *above* docsets that declare bundle rules, `ValidateBundle` returns those docset roots and `lore validate` prints `lore validate: / is above docsets docs, wiki with bundle rules; run lore validate per docset` to stderr instead of a silent pass. Acceptance test added to Phase 1 ("Bundle anchor").

pkg/rules/okfrule/okf.go
```diff
@@ -0,0 +1,60 @@
...
+				if okf.IsReserved(subject.Path) {
+					return nil, nil
+				}
+				return []rules.Finding{{Code: "okf/concept", Measured: err.Error()}}, nil
+			}
+			return nil, err
+		}
+		return nil, nil
+	}
+	if subject.Bundle == nil || len(subject.Bundle.Files) == 0 || subject.Path != subject.Bundle.Files[0].AbsolutePath {
```
aakarim: Part of blocking (1): this `Files[0]` guard moves into the engine — `Evaluate` for a bundle member should be called exactly once per bundle by `ValidateBundle`, with `Subject.Path = bundle root`. Same at `link/link.go:43`.

pkg/openlore/server.go
```diff
@@ -256,10 +257,19 @@ func newServerWithRoot(rootDir string, rootFS, lowerFS vfs.FileSystem, opts ...c
 			return nil, err
 		}
 	}
-	if anyDocsetHasOKF(s.auth.Docsets) {
-		if err := s.registerPlugin(newOKF(s.auth.Docsets, logger)); err != nil {
+	if anyConfiguredRules(s.auth) {
```
aakarim, line 260: **Blocking (2): `okfPlugin` admission/validation is now dead code but still present in full.**

`okfPlugin` is no longer registered here, only its `MetaExtenders()` are kept, yet `okf_plugin.go` still has the complete `WriteMiddleware()` and `Validators()` including a second implementation of the broken-link / link-outside-bundle / alias-referrer / alias-target checks that now live in `pkg/rules/link`. Two copies that must stay in sync is exactly what the rules engine was meant to remove. `okf_plugin_test.go` "passes unmodified" only because it tests `newOKF` directly, not the server.

Plan §3.4: `okfPlugin.Validators()` becomes a thin shim over the engine (so existing callers/tests keep working) and its `WriteMiddleware` and link/alias code are deleted. Retarget the tests that assert admission behaviour at the server or at `rulesPlugin`. Also: the boot log now says `name=rules version=0.1.0`; `plugin_info_test` should cover the plugin that is actually registered.

pkg/openlore/rules_plugin_test.go
```diff
@@ -0,0 +1,101 @@
...
+	"strings"
+	"testing"
+
+	"github.com/aakarim/go-openlore/internal/config"
+	"github.com/aakarim/go-openlore/pkg/openlore/validation"
+	"github.com/aakarim/go-openlore/pkg/rules"
+	"github.com/aakarim/go-openlore/pkg/vfs"
+)
+
+func newTestRulesPlugin(t *testing.T, auth *config.AuthConfig, logger *slog.Logger) *rulesPlugin {
```
aakarim, line 19: **Blocking (3): no test through the real boot path.**

Every test here builds `newRulesPlugin` directly. Phase 1 acceptance tests 1, 3, 4, 5 and 7 are specified as *server* tests (real `NewServer(root, WithReadonly(false), config.WithDataDir, config.WithAuthFile)` + writes through `s.writeChain()` + `lore validate` via the shell). Blocking (1) would have been caught by one. For reference, a probe I ran that way confirmed: legacy `okf` block still rejects with the `okf: <path>: …` prefix, a global `rules:` entry applies to a docset with no rules of its own, the `size/lines` block matches the plan text, `len(s.validators)==1`. Please add those as real tests plus the new "Bundle anchor" case from the plan.

pkg/rules/engine.go
```diff
@@ -0,0 +1,263 @@
...
+		layers = append(layers, got...)
+	}
+	if len(layers) == 0 {
+		return nil, nil
+	}
+	specs, origins, err := Unify(layers)
+	if err != nil {
+		return nil, err
+	}
+	scope := layers[len(layers)-1].Scope
```
aakarim: **Design (4): scope must belong to each rule's declaring layer, not the innermost layer.**

All unified rules are compiled with `layers[last].Scope`. Harmless today because every lore.json layer shares the docset root, but wrong the moment Phase 3 adds folder layers: a global `match: ["adr/**"]` tightened by `/docs/backend/.lore/config.yaml` would be evaluated as `/docs/backend/adr/**`. Inner layers may tighten `with` but never move globs, so the scope is that of the outermost origin.

Plan §4.4 updated: `Unify(layers) (map[string]UnifiedRule, error)` where `UnifiedRule{Spec, Origins, Scope}`, and `Compile(reg, env, unified)` with no `scope` argument. `CompiledRule.Scope` comes from `UnifiedRule.Scope`.

(Not asking for a cache here: I benchmarked `Effective` on this plugin's engine with five rules at ≈4.5 µs / 47 allocs per call. The plan comment on `Member.Compile` now says compile is cheap/pure and may run per write; only Phase 3 folder-layer *reads* get cached.)

pkg/rules/okfrule/okf.go
```diff
@@ -0,0 +1,60 @@
...
+
+func (c check) Evaluate(_ context.Context, subject rules.Subject) ([]rules.Finding, error) {
+	if !c.bundle {
+		if err := okf.Validate(subject.Path, subject.Content); err != nil {
+			if subject.Mode == rules.ModeValidate {
+				if okf.IsReserved(subject.Path) {
+					return nil, nil
+				}
+				return []rules.Finding{{Code: "okf/concept", Measured: err.Error()}}, nil
+			}
```
aakarim, line 39: **Design (5): OKF rejection should be legacy line + `see:` line.**

Returning the `okf.Validate` error from `Evaluate` in `ModeAdmit` routes it through the fail-closed path, so OKF writes get the bare `okf: /docs/bad.md: missing YAML frontmatter block …` with no `see:` line, while `size/*` gets the full block. Agreed resolution (plan §3.6): keep the legacy first line verbatim for compatibility and append exactly one line, `  see: lore package doc okf` — nothing else, since the `pkg/okf` message already names the fix. Return it as a `Finding` (`Measured = err.Error()`) and let the engine special-case the `okf:` prefix, or give `Rejection` a minimal-format flag; either way `Evaluate` returning an error should mean "member failed", not "file non-conformant". Phase 1 acceptance test 1 now asserts "legacy line followed by exactly one more line".

pkg/rules/engine.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+	return strings.Join(parts, "\n")
+}
+
+func (r *Rejection) Unwrap() error { return r.Err }
+
+func findingText(f Finding, rule CompiledRule, target string) string {
+	limit := f.Limit
+	if limit == "" {
+		limit = "this write cannot proceed under this rule"
```
aakarim: **Design (6): validate-time output must differ from write-time output.**

Today `lore validate` prints the full write-time block under each diagnostic, including `this write cannot proceed under this rule` (wrong — nothing is being written) and the placeholder `suggested: change the file so it satisfies the rule`. Plan §3.6 (updated): the four-question block is write-only. `lore validate` prints the existing one-line diagnostic (`path:line:col: severity [rule] message`, message = line 1 of the block) with no `suggested:`/`override:` lines, and after the listing one `see: lore package doc <member>` line per distinct member that produced findings. Drop the generic fallbacks for `Limit`/`Remedy` entirely — an empty line is better than filler that costs the agent tokens and says nothing.

pkg/openlore/rules_plugin.go
```diff
@@ -0,0 +1,109 @@
...
+			}
+		}
+	}
+	return bestRoot, bestName, best, bestLen >= 0
+}
+
+type rulesPlugin struct{ engine *rules.Engine }
+
+func newRulesPlugin(auth *config.AuthConfig, defaults rules.Defaults, logger *slog.Logger) (*rulesPlugin, error) {
+	registry := rules.NewRegistry()
```
aakarim: **Design (7): two registries.**

`size`, `okfrule` and `link` all `init()`-register into `rules.DefaultRegistry()` (per plan §4.3), but the plugin builds a private `NewRegistry()` because `link/alias` needs the docset alias roots. Result: two registries whose `link/alias` members differ, and Phase 2 `lore package list/doc` would have to know which one the engine uses. Plan §4.3 updated: `Env` gains `AliasRoots []string` (longest first); `link.Member{Alias: true}` reads them in `Compile(with, env)`. Members are then config-free values and `DefaultRegistry` is the only registry — pass it in `rules.Options.Registry` here.

internal/config/config.go
```diff
@@ -1368,3 +1395,64 @@ func ValidateAuthConfig(auth *AuthConfig) error {
...
+				return fmt.Errorf("docset %q has conflicting okf and rules.okf configuration", name)
+			}
+		} else {
+			docset.Rules["okf"] = expected
+		}
+		if _, ok := docset.Rules["okf/bundle"]; !ok {
+			docset.Rules["okf/bundle"] = rules.RuleSpec{Match: []string{"**/*.md"}, Use: "okf/bundle", Enforce: docset.OKF.Enforce}
+		}
+		if _, ok := docset.Rules["link/resolves"]; !ok {
+			docset.Rules["link/resolves"] = rules.RuleSpec{Match: []string{"**/*.md"}, Use: "link/resolves", Enforce: docset.OKF.Enforce}
```
aakarim, line 1427: Minor (8): `okf/bundle` and `link/resolves` inheriting the `okf` block's `enforce` means a docset with `okf: {enforce: false}` now reports broken links and index-structure problems as warnings where main always reported errors. Agreed this is the right behaviour (one switch for the OKF family) — plan §2.2 now says so. Please call it out in the PR description as a behaviour change.

internal/config/config.go
```diff
@@ -1368,3 +1395,64 @@ func ValidateAuthConfig(auth *AuthConfig) error {
...
+		if _, ok := docset.Rules["link/alias"]; !ok {
+			warn := false
+			docset.Rules["link/alias"] = rules.RuleSpec{Match: []string{"**/*.md"}, Use: "link/alias", Enforce: &warn}
+		}
+		auth.Docsets[name] = docset
+	}
+	return nil
+}
+
+func sameLegacyOKFRule(a, b rules.RuleSpec) bool {
```
aakarim: Minor (10): `sameLegacyOKFRule` re-implements the field-by-field equality that `rules.differingKeys` already does for unification. Export a `RuleSpec.Equal` (or `DifferingKeys`) in `pkg/rules` and use it here so the two cannot drift.

pkg/rules/types.go
```diff
@@ -0,0 +1,135 @@
...
+	ModeAdmit Mode = iota
+	ModeValidate
+)
+
+type Finding struct {
+	Code     string
+	Path     string
+	Line     int
+	Column   int
+	Warning  bool
```
aakarim, line 108: Minor (9): `Finding.Warning` was not in the plan but is the right way to carry `pkg/okf`'s warning-severity diagnostics. Added to the plan (§4.3) with these semantics: never rejects a write; reports as `warning` in validate even when the rule enforces; `enforce` decides error vs warning for everything else. Please add that as the field comment.

pkg/rules/okfrule/okf.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+func (m Member) Compile(_ map[string]any, _ rules.Env) (rules.Check, error) {
+	return check{bundle: m.Bundle}, nil
+}
+
+type check struct{ bundle bool }
+
+func (c check) Evaluate(_ context.Context, subject rules.Subject) ([]rules.Finding, error) {
+	if !c.bundle {
```
aakarim, line 32: Minor (10): with a custom `okf.patterns` (e.g. `*.txt`), the desugared file rule now runs `okf.Validate` on non-`.md` matches in `ModeValidate`, whereas main's `ValidateBundle` skipped non-`.md` files. Either keep the `.md` guard for validate or make it explicit in the plan that patterns govern validate too — I'd keep parity with main for this phase.

aakarim (commented):
Reviewed against `docs/rules-system.md` (plan updated in the same pass; section refs below point at the updated plan). Engine core — matcher, unify, compile, size messages, tokenizer, did-you-mean errors — matches the plan and is good. Three things block merge; the rest are design alignments and minor cleanups, all agreed with Adil.

**Blocking**
1. Bundle-scope rules are anchored on `Files[0]` instead of the bundle root (`engine.go`, `okfrule/okf.go`, `link/link.go`).
2. `okfPlugin` still carries its full `WriteMiddleware`/`Validators` (with a second copy of the link/alias checks) although `server.go` no longer registers it for admission — dead code that must stay in sync with `pkg/rules/link`.
3. No test exercises the real `NewServer` boot path; all plugin tests construct `newRulesPlugin` directly.

**Design alignment (plan §3.4, §3.6, §4.3, §4.4)**
4. Per-rule scope instead of innermost-layer scope.
5. OKF rejection message = legacy line + `see:` line only.
6. `lore validate` output must be one-line diagnostics + one trailing `see:` per member, no `suggested:`/`override:` lines.
7. Single registry: alias roots via `Env`, drop the private `NewRegistry()`.

**Minor**
8. Desugared `enforce` inheritance — fine, add to PR description and the plan already records it.
9. `Finding.Warning` — fine, now in the plan.
10. `sameLegacyOKFRule` duplicates `differingKeys`; boot log name; custom OKF patterns on non-`.md` files.

Verified locally: `go build ./...`, `go vet`, `go test ./pkg/rules/... ./internal/config/ ./pkg/openlore/` all pass. `Effective` benchmarks at ≈4.5 µs / 47 allocs for five rules, so no compiled-rule cache is needed (plan comment on `Member.Compile` updated to say so).

PR #67, opened 2026-09-03
Rules: add Phase 2 package discoverability

aakarim:
## Summary
- add `lore package list` and `lore package doc <path>` backed by rule manifests
- generate the standard-library rule reference and enforce drift checks in CI
- document folder-rule discovery in the bundled OpenLore skill

## Verification
- `go test -race ./...`
- `go build ./...`
- `go generate ./docs/... && git diff --exit-code docs/rules-stdlib.md`


pkg/rules/size/size.go
```diff
@@ -30,16 +30,22 @@ func Register(reg *rules.Registry) {
...
+		memberPath, summary = "size/tokens", "Reject writes with more than max tokens"
+		doc = "Reject writes whose estimated token count exceeds max. Tokens use estimate/v1: ceil(bytes / 4)."
 	}
-	return rules.Manifest{Path: path, Kind: rules.KindRule, Scope: rules.ScopeFile, Summary: summary,
-		Params:  []rules.Param{{Name: "max", Type: rules.ParamIntegerOrInitial, Required: true}, {Name: "growth", Type: rules.ParamNumber}},
-		Example: "use: " + path + "\nwith: { max: 60 }"}
+	return rules.Manifest{Path: memberPath, Kind: rules.KindRule, Scope: rules.ScopeFile, Summary: summary, Doc: doc,
+		Params: []rules.Param{
+			{Name: "max", Type: rules.ParamIntegerOrInitial, Required: true, Doc: "Fixed cap; max: initial ships in folder rules phase 4."},
+			{Name: "growth", Type: rules.ParamNumber, Doc: "Multiplier for max: initial; defaults to openlore.yml rules.growth (1.25)."},
```
aakarim: **Decision (Adil): don't advertise `max: initial` until Phase 4 ships it.** An agent reading `integer | "initial"` here will try it, hit the compile error, and learn nothing actionable; and "phase 4" is our roadmap vocabulary leaking into agent-facing output (this text lands in the shell, `docs/rules-stdlib.md`, and the SKILL flow).

For this PR: declare only `max` with `Type: rules.ParamInteger, Required: true, Doc: "Fixed cap."` and drop the `growth` param. Phase 4 adds `initial`/`growth` back together with the behaviour. `displayParamType` in `docs.go` can then go too (or stay, it's harmless). I'll mark the plan's §4.9 `max: initial, growth: 1.1` example as Phase 4.

Same for the `max: initial is not supported until folder rules phase 4` special case in `Compile` (line 53, outside this hunk): once `initial` is not a declared type, drop it and let `max must be an integer >= 1` be the error — no roadmap reference in messages an agent will read.

pkg/shell/cmds/lore_package.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+func cmdLorePackage(_ CmdContext, args []string, w io.Writer, errW io.Writer, _ io.Reader) int {
+	if len(args) == 1 && args[0] == "list" {
+		rules.WriteList(w, rules.DefaultRegistry())
+		return 0
+	}
+	if len(args) == 2 && args[0] == "doc" {
+		name := args[1]
+		if member, ok := rules.DefaultRegistry().Lookup(name); ok {
```
aakarim: **Decision (Adil): when a path is both a member and a package, show both.** `lore package doc okf` currently prints the `okf` rule and never mentions `okf/bundle`, so the only way to discover it is `list`. After `WriteDoc`, if `PackageMembers(reg, name)` is non-empty, append a `MEMBERS` section listing them (same row format as the package listing below), like `go doc` showing a package's exports. Worth a case in `TestLorePackageDoc`: `lore package doc okf` contains `okf/bundle`.

pkg/shell/cmds/lore_package.go
```diff
@@ -0,0 +1,44 @@
...
+			rules.WriteDoc(w, member.Manifest())
+			return 0
+		}
+		if members := rules.PackageMembers(rules.DefaultRegistry(), name); len(members) != 0 {
+			for _, member := range members {
+				manifest := member.Manifest()
+				fmt.Fprintf(w, "%s\t%s\t%s\t%s\n", manifest.Path, manifest.Kind, manifest.Scope, manifest.Summary)
```
aakarim: This prints raw tab-separated rows with no header, while `list` uses tabwriter with a `NAME KIND SCOPE SUMMARY` header — `lore package doc size` and `lore package list` should look the same. Simplest: have `WriteList` take the members (or a filter) and call it from both paths.

pkg/rules/docs.go
```diff
@@ -0,0 +1,72 @@
...
+	if len(manifest.Params) != 0 {
+		fmt.Fprintln(w, "\nPARAMETERS")
+		for _, parameter := range manifest.Params {
+			setting := "optional"
+			if parameter.Required {
+				setting = "required"
+			} else if parameter.Default != nil {
+				setting = fmt.Sprintf("default: %v", parameter.Default)
+			}
+			fmt.Fprintf(w, "  %-10s %-17s %-12s %s\n", parameter.Name, displayParamType(parameter.Type), setting, parameter.Doc)
```
aakarim: `%-17s` is narrower than the 19-char `integer | "initial"`, so the PARAMETERS table misaligns in both the shell and the generated `docs/rules-stdlib.md` (see `max` vs `growth` rows). With the `initial` type gone this stops biting today, but a `tabwriter` here would keep it from coming back in Phase 4 — `WriteList` already uses one.

aakarim (commented):
Reviewed against the plan's Phase 2 (`docs/rules-system.md` §4.9 / Phase 2 acceptance tests). Verified locally on `d86cc01`: `go vet ./...`, `go test ./...`, `gofmt`, and `go generate ./docs/... && git diff --exit-code docs/rules-stdlib.md` are all clean; `lore package list` prints all seven members and `doc` output matches the plan's shape. Acceptance tests 1, 2 and 4 are covered here; 3 and 5 were already covered by Phase 1 tests.

Two changes below come from design decisions Adil made for this PR (agent-facing docs must not advertise `max: initial` until Phase 4; `lore package doc <pkg>` must show sub-members when the path is also a member). The rest are small.

pkg/rules/docs.go
```diff
@@ -0,0 +1,72 @@
...
+	} else {
+		fmt.Fprintln(w, manifest.Summary+".")
+	}
+	if len(manifest.Params) != 0 {
+		fmt.Fprintln(w, "\nPARAMETERS")
+		for _, parameter := range manifest.Params {
+			setting := "optional"
+			if parameter.Required {
+				setting = "required"
+			} else if parameter.Default != nil {
```
aakarim, line 55: Minor: no member sets `Param.Default`, so this branch is dead for now. Fine to keep — just noting that `growth`'s default is config-driven (`openlore.yml rules.growth`), so when it returns in Phase 4 it correctly belongs in `Doc`, not `Default`.

pkg/shell/cmds/lore_package_test.go
```diff
@@ -0,0 +1,54 @@
...
+func TestLorePackageList(t *testing.T) {
+	out, errOut, code := runMeta(t, shell.NewShell(testFS()), "/", "lore package list")
+	if code != 0 || errOut != "" {
+		t.Fatalf("exit=%d stderr=%q", code, errOut)
+	}
+	lines := strings.Split(strings.TrimSpace(out), "\n")
+	if len(lines) != 8 {
+		t.Fatalf("got %d lines, want header plus 7 members:\n%s", len(lines), out)
+	}
+	for i, name := range []string{"link/alias", "link/resolves", "okf", "okf/bundle", "size/kilobytes", "size/lines", "size/tokens"} {
```
aakarim: Hardcoding the seven names means this test churns every time a member is added or renamed. Derive the expectation from `rules.DefaultRegistry().All()` (sorted by `Manifest().Path`) and assert the rows match one-to-one; that still checks "exactly the registered members, sorted" per acceptance test 1 without duplicating the list. Keep one literal spot-check (e.g. `size/lines` row has `rule` and `file`) so a broken registry can't make the test vacuously pass.

docs/rules-stdlib.md
```diff
@@ -0,0 +1,90 @@
+<!-- Code generated by go generate; DO NOT EDIT. -->
+
+# Rules standard library
```
aakarim, line 3: Nothing links here yet. Please add a row to the Documentation table in `README.md` (`| [Rules standard library](docs/rules-stdlib.md) | Generated reference for compiled-in rule members and their parameters |`) and a `lore package list` / `lore package doc <path>` row to the `lore` table in `docs/commands.md` next to `lore validate` — that table is where agents and people look for shell commands.

pkg/rules/size/size.go
aakarim: Addressed in ec237af: size manifests now expose only a required integer max, and the Phase 4-only compile branches and roadmap-facing text are removed.

pkg/shell/cmds/lore_package.go
aakarim: Addressed in ec237af: member docs now append a MEMBERS table when the same path is also a package; the okf/okf/bundle case is covered by a test.

pkg/shell/cmds/lore_package.go
aakarim: Addressed in ec237af: WriteList now accepts a member slice and renders both the global and package-filtered listings with the same header and tabwriter formatting.

pkg/rules/docs.go
aakarim: Addressed in ec237af: parameter rows now use tabwriter, so future wider types remain aligned.

pkg/rules/docs.go
aakarim: Noted. The config-driven default branch remains generic renderer support; the growth parameter and its default text are removed until the behavior ships.

pkg/shell/cmds/lore_package_test.go
aakarim: Addressed in ec237af: list expectations are derived one-to-one from DefaultRegistry().All(), with a separate literal size/lines rule/file row check.

docs/rules-stdlib.md
aakarim: Addressed in ec237af: the generated reference is linked from README.md and both package discovery commands are documented in docs/commands.md.

PR #69, opened 2026-09-03
Implement Phase 3 folder rule configuration

aakarim:
## Summary
- expose only `.lore/config.yaml` through DirFS while keeping package state hidden
- strictly decode and layer folder rules with cache invalidation on config changes
- enforce per-docset `config.edit` permissions and exempt config files from content rules
- report folder config errors through `lore validate`
- cover VFS visibility, permissions, inheritance, invalidation, unification, validation, and shell visibility

## Testing
- `go test ./...`
- `go test -race ./pkg/openlore ./pkg/rules -run "(FolderRules|DirFSExposesOnlyFolderConfig|DecodeFile)"`

### DIFF
--- a/pkg/openlore/authz.go
+++ b/pkg/openlore/authz.go
@@ -2,6 +2,7 @@ package openlore
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io/fs"
 	"path"
@@ -416,6 +417,9 @@ func (s *Server) identityCanWrite(id Identity, action vfs.ChangeAction, p string
 		return false
 	}
 	if action == vfs.ChangeActionRemoveAll {
+		if !s.canEditDirConfigsInTree(id, p) {
+			return false
+		}
 		for _, candidate := range s.currentAuth().Docsets {
 			for _, pm := range candidate.Paths {
 				root := displayPath(pm)
@@ -448,6 +452,23 @@ func (s *Server) identityCanWrite(id Identity, action vfs.ChangeAction, p string
 	return false
 }
 
+func (s *Server) canEditDirConfigsInTree(id Identity, target string) bool {
+	allowed := true
+	err := vfs.WalkDir(s.merge, target, func(candidate string, _ *vfs.FileInfo, walkErr error) error {
+		if walkErr != nil {
+			return walkErr
+		}
+		if isDirConfigPath(candidate) && !s.identityHasDirConfigRole(id, candidate) {
+			allowed = false
+		}
+		return nil
+	})
+	if errors.Is(err, fs.ErrNotExist) {
+		return true
+	}
+	return err == nil && allowed
+}
+
 func (s *Server) identityHasDirConfigRole(id Identity, target string) bool {
 	_, docset, ok := s.mostSpecificDocset(path.Dir(path.Dir(target)))
 	if !ok || docset.Config == nil {
@@ -465,21 +486,6 @@ func (s *Server) identityHasDirConfigRole(id Identity, target string) bool {
 		}
 	}
 	return false
-}
-
-// CanEditDirConfig applies the additional role gate for .lore/config.yaml.
-// Ordinary path authorization remains required as well.
-func (s *Server) CanEditDirConfig(attribution Attribution, target string) bool {
-	id := Identity{
-		IdentityName: attribution.Principal,
-		Principal:    AuthenticatedPrincipal{IdentityName: attribution.Principal},
-		Attribution:  attribution,
-		Scopes:       []string{ScopeFull},
-	}
-	if !s.identityCanWrite(id, vfs.ChangeActionWrite, target) {
-		return false
-	}
-	return true
 }
 
 // readableRoots returns the display roots the identity may read: the display
--- a/pkg/openlore/authz.go
+++ b/pkg/openlore/authz.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io/fs"
+	"path"
 	"sort"
 	"strings"
 	"syscall"
@@ -411,6 +412,9 @@ func (s *Server) identityCanWrite(id Identity, action vfs.ChangeAction, p string
 		return false // token scope ceiling: only full authority may write
 	}
 	p = s.canonicalPath(p)
+	if isDirConfigPath(p) && !s.identityHasDirConfigRole(id, p) {
+		return false
+	}
 	if action == vfs.ChangeActionRemoveAll {
 		for _, candidate := range s.currentAuth().Docsets {
 			for _, pm := range candidate.Paths {
@@ -442,6 +446,40 @@ func (s *Server) identityCanWrite(id Identity, action vfs.ChangeAction, p string
 		}
 	}
 	return false
+}
+
+func (s *Server) identityHasDirConfigRole(id Identity, target string) bool {
+	_, docset, ok := s.mostSpecificDocset(path.Dir(path.Dir(target)))
+	if !ok || docset.Config == nil {
+		return false
+	}
+	policy, err := s.currentPolicy(id)
+	if err != nil {
+		return false
+	}
+	for _, allowed := range docset.Config.Edit {
+		for _, role := range policy.Roles {
+			if role == allowed {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// CanEditDirConfig applies the additional role gate for .lore/config.yaml.
+// Ordinary path authorization remains required as well.
+func (s *Server) CanEditDirConfig(attribution Attribution, target string) bool {
+	id := Identity{
+		IdentityName: attribution.Principal,
+		Principal:    AuthenticatedPrincipal{IdentityName: attribution.Principal},
+		Attribution:  attribution,
+		Scopes:       []string{ScopeFull},
+	}
+	if !s.identityCanWrite(id, vfs.ChangeActionWrite, target) {
+		return false
+	}
+	return true
 }
 
 // readableRoots returns the display roots the identity may read: the display
--- a/pkg/openlore/rules_plugin.go
+++ b/pkg/openlore/rules_plugin.go
@@ -42,15 +42,7 @@ func newFolderRuleLayers(fsys vfs.FileSystem, docsets map[string]config.DocsetSp
 	return &folderRuleLayers{fs: fsys, docsets: docsets, cache: map[string]folderRuleCacheEntry{}}
 }
 
-func (s *folderRuleLayers) LayersFor(ctx context.Context, target string) ([]rules.Layer, error) {
-	dir := path.Dir(vfs.CleanPath(target))
-	if rules.IsDirConfigPath(target) {
-		dir = path.Dir(path.Dir(vfs.CleanPath(target)))
-	} else if s.fs != nil {
-		if info, err := s.fs.Stat(target); err == nil && info.Dir {
-			dir = vfs.CleanPath(target)
-		}
-	}
+func (s *folderRuleLayers) LayersForDir(ctx context.Context, dir string) ([]rules.Layer, error) {
 	return s.layersThrough(ctx, dir, true)
 }
 
@@ -197,17 +189,11 @@ func ruleRoots(docset config.DocsetSpec) []string {
 	return roots
 }
 
-type dirConfigAuthorizer interface {
-	CanEditDirConfig(Attribution, string) bool
-}
-
 type rulesPlugin struct {
-	engine     *rules.Engine
-	authorizer dirConfigAuthorizer
-	reportInfo bool
+	engine *rules.Engine
 }
 
-func newRulesPlugin(auth *config.AuthConfig, defaults rules.Defaults, fsys vfs.FileSystem, authorizer dirConfigAuthorizer, logger *slog.Logger) (*rulesPlugin, error) {
+func newRulesPlugin(auth *config.AuthConfig, defaults rules.Defaults, fsys vfs.FileSystem, logger *slog.Logger) (*rulesPlugin, error) {
 	var aliases []string
 	for _, docset := range auth.Docsets {
 		aliases = append(aliases, docset.Aliases...)
@@ -231,11 +217,7 @@ func newRulesPlugin(auth *config.AuthConfig, defaults rules.Defaults, fsys vfs.F
 		options.Folders = newFolderRuleLayers(fsys, auth.Docsets)
 	}
 	engine := rules.New(options)
-	reportInfo := anyConfiguredRules(auth)
-	for _, docset := range auth.Docsets {
-		reportInfo = reportInfo || docset.Config != nil
-	}
-	return &rulesPlugin{engine: engine, authorizer: authorizer, reportInfo: reportInfo}, nil
+	return &rulesPlugin{engine: engine}, nil
 }
 
 func (p *rulesPlugin) WriteMiddleware() []WriteMiddleware {
@@ -243,9 +225,6 @@ func (p *rulesPlugin) WriteMiddleware() []WriteMiddleware {
 		return func(ctx context.Context, op WriteOp) (WriteResult, error) {
 			for _, leaf := range op.Leaves() {
 				if rules.IsDirConfigPath(leaf.Target) {
-					if p.authorizer == nil || !p.authorizer.CanEditDirConfig(op.Attribution, leaf.Target) {
-						return WriteResult{}, os.ErrPermission
-					}
 					if leaf.Action == vfs.ChangeActionWrite && leaf.Write != nil {
 						dir := path.Dir(path.Dir(leaf.Target))
 						if err := p.engine.CheckConfigFile(ctx, dir, leaf.Write.Bytes); err != nil {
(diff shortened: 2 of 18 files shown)


aakarim (commented):
Reviewed against `docs/rules-system.md` Phase 3 (§3.5, §3.8, acceptance tests 0–8). Verified on `593f31e`: `go vet ./...`, `go test ./...` and gofmt are clean, and the new tests cover every acceptance test in the list. Nice work on the DirFS carve-out and the unification/`default: true` coverage.

One item comes from a design decision Adil made for this PR: deleting a folder that contains a `.lore/config.yaml` must require `config.edit` (`RemoveAll` currently bypasses the gate — I verified an `rw`-only identity can delete `/docs/sub` and its config on the branch). I've probed two other behaviours below (nested `.lore` path escapes the reserved check; error types differ per entry point). The rest are small.

pkg/openlore/authz.go
```diff
@@ -411,6 +412,9 @@ func (s *Server) identityCanWrite(id Identity, action vfs.ChangeAction, p string
 		return false // token scope ceiling: only full authority may write
 	}
 	p = s.canonicalPath(p)
+	if isDirConfigPath(p) && !s.identityHasDirConfigRole(id, p) {
+		return false
+	}
 	if action == vfs.ChangeActionRemoveAll {
```
aakarim, line 419: **Decision (Adil): `RemoveAll` of a tree containing any `.lore/config.yaml` requires `config.edit`.** Today the gate on line 415 only fires when the leaf *is* the config file. Probed on this branch: bob (`rw`, no `config.edit`) gets `os.ErrPermission` writing or removing `/docs/sub/.lore/config.yaml`, but `RemoveAll /docs/sub` succeeds and deletes the config with the folder — so a non-editor can drop governance by deleting and recreating the folder.

In this `RemoveAll` branch, walk the target on the host (the snapshot already lists `.lore/config.yaml` entries after your `rawSnapshot` change) and return false unless `identityHasDirConfigRole` holds for each config found. Please add a server test: bob's `RemoveAll /docs/sub` → `os.ErrPermission`; alice's succeeds; bob's `RemoveAll` of a folder with no config still succeeds. I'll add this to plan §3.5 and the Phase 3 acceptance tests.

pkg/openlore/vfs.go
```diff
@@ -280,6 +282,11 @@ func hasReservedPath(p string) bool {
 	return false
 }
 
+func isDirConfigPath(p string) bool {
+	clean := vfs.CleanPath(p)
+	return path.Base(clean) == "config.yaml" && path.Base(path.Dir(clean)) == ".lore"
```
aakarim: This only inspects the last two segments, so `/docs/.lore/xattrs/.lore/config.yaml` satisfies `hasReservedPath(p) && !isDirConfigPath(p) == false` and gets through every gate. Verified on this branch: alice's `WriteFileAtomic` on that path succeeds and creates the file inside the host xattrs store. Require that the grandparent directory has no `.lore` component (i.e. exactly one `.lore`, in the penultimate position). Same fix for `rules.IsDirConfigPath` in `pkg/rules/engine.go` — ideally the plugin calls the one in `pkg/rules` so the two can't drift.

pkg/openlore/scoped_write_fs.go
```diff
@@ -36,6 +38,9 @@ func newScopedWriteFS(base vfs.FileSystem, authz writeAuthorizer) *scopedWriteFS
 
 func (s *scopedWriteFS) WriteFileAtomic(p string, data []byte, opts vfs.WriteOpts) (string, error) {
 	if s.inner == nil || !s.authorize(vfs.ChangeActionWrite, p) {
```
aakarim, line 40: Only `WriteFileAtomic` and `AdmitChangeSet` map a denied config path to `os.ErrPermission`; `Remove` and `RemoveAll` below still return `vfs.ErrReadOnly` (verified: bob's `Remove` of the config reports "read-only filesystem"), and `Server.AdmitChangeSet` in `server.go` returns `ErrReadOnly` too. §3.5 says `os.ErrPermission` for every denied config write. Simplest: have `identityCanWrite` callers ask a tiny helper `denyError(p)` that picks the error, and use it in all five places.

pkg/openlore/authz.go
```diff
@@ -444,6 +448,40 @@ func (s *Server) identityCanWrite(id Identity, action vfs.ChangeAction, p string
...
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// CanEditDirConfig applies the additional role gate for .lore/config.yaml.
+// Ordinary path authorization remains required as well.
+func (s *Server) CanEditDirConfig(attribution Attribution, target string) bool {
```
aakarim: This rebuilds an `Identity` from `Attribution.Principal` with a hard-coded `ScopeFull` and re-runs `identityCanWrite`. Both write entry points already ran `identityCanWrite` with the *real* identity before the middleware sees the op — `buildSessionFS` → `scopedWriteFS`, and `Server.AdmitChangeSet` — and both now include the `config.edit` check via line 415. So this second gate is redundant, and the reconstructed identity is weaker than the original (no token scope, no delegated-actor context). Unless there is a write path that reaches `writeChain()` without one of those two, I'd drop `CanEditDirConfig`/`dirConfigAuthorizer` and keep the middleware to `CheckConfigFile` + invalidation only. If there is such a path, please name it in a comment here.

pkg/openlore/rules_plugin.go
```diff
@@ -77,44 +197,103 @@ func ruleRoots(docset config.DocsetSpec) []string {
...
+					continue
+				}
 				if err := p.engine.AdmitLeaf(ctx, leaf, op.Attribution.String(), nil); err != nil {
 					return WriteResult{}, err
 				}
 			}
-			return next(ctx, op)
+			result, err := next(ctx, op)
+			if err == nil {
+				p.invalidateChanges(op.Leaves())
```
aakarim: The cache is invalidated twice: here, right after `next` returns, and again in `PostCommitMiddleware`. The plan (§4.4, Phase 3 scope) says post-commit. If `next` returns before the change is visible in the FS (queued/inbox writes), this early invalidation is a no-op that lets the next `load` re-cache the *old* content — so it's not just redundant, it can pin a stale entry until the post-commit hook runs. Keep only the post-commit one.

pkg/openlore/rules_plugin.go
```diff
@@ -77,44 +197,103 @@ func ruleRoots(docset config.DocsetSpec) []string {
...
+			return next(ctx, info)
 		}
 	}}
 }
 func (p *rulesPlugin) Validators() []validation.Validator {
 	return []validation.Validator{p.engine.Validator()}
 }
-func (p *rulesPlugin) Info() PluginInfo { return PluginInfo{Name: "rules", Version: "0.1.0"} }
+func (p *rulesPlugin) Info() PluginInfo {
+	if !p.reportInfo {
```
aakarim: Returning an empty `PluginInfo` to hide the plugin (plus the `info.Name != ""` special case in `server.go`) means a host-placed `.lore/config.yaml` can be enforcing rules while `rules` is absent from the plugin list and the boot log — exactly the situation where an operator would look there. Now that the plugin is always registered, always report `rules 0.1.0` and drop `reportInfo`.

pkg/rules/engine.go
```diff
@@ -37,6 +37,101 @@ func (e *BundleRootError) Error() string {
...
+	var layers []Layer
+	if e.options.Config != nil {
+		got, err := e.options.Config.LayersFor(ctx, target)
+		if err != nil {
+			return fmt.Errorf("%s: %w", configPath, err)
+		}
+		layers = append(layers, got...)
+	}
+	if e.options.Folders != nil {
+		folders, ok := e.options.Folders.(FolderLayerSource)
```
aakarim: `Options.Folders` is a `LayerSource` and both `CheckConfigFile` and `Invalidate` type-assert to `FolderLayerSource` (with an error path here that can't be exercised by any real wiring). Type the field as `FolderLayerSource` and the assertions and the "cannot validate config" branch go away.

pkg/openlore/rules_plugin.go
```diff
@@ -21,6 +25,122 @@ type configRuleLayers struct {
...
+func newFolderRuleLayers(fsys vfs.FileSystem, docsets map[string]config.DocsetSpec) *folderRuleLayers {
+	return &folderRuleLayers{fs: fsys, docsets: docsets, cache: map[string]folderRuleCacheEntry{}}
+}
+
+func (s *folderRuleLayers) LayersFor(ctx context.Context, target string) ([]rules.Layer, error) {
+	dir := path.Dir(vfs.CleanPath(target))
+	if rules.IsDirConfigPath(target) {
+		dir = path.Dir(path.Dir(vfs.CleanPath(target)))
+	} else if s.fs != nil {
+		if info, err := s.fs.Stat(target); err == nil && info.Dir {
```
aakarim: This does a `Stat` on every `LayersFor` call, i.e. on every admitted leaf, to guess whether the target is a directory. Callers know: `AdmitLeaf` passes file targets, `Effective(bundle.Root)` passes a directory. Give the engine a `LayersForDir(dir)` (or pass the directory in from `Effective`) rather than probing the FS from the layer source.

pkg/openlore/rules_plugin.go
```diff
@@ -21,6 +25,122 @@ type configRuleLayers struct {
...
+			continue
+		}
+		current = path.Join(current, segment)
+		result = append(result, current)
+	}
+	return result
+}
+
+func (s *folderRuleLayers) load(dir string) folderRuleCacheEntry {
+	s.mu.Lock()
```
aakarim, line 100: Minor: this takes the exclusive lock for the whole `ReadFile` + decode, so every admission on the server serialises behind one folder's IO on a cache miss. Fine for Phase 3; a read-lock fast path (check cache under `RLock`, upgrade only on miss) is a two-line change if it ever shows up.

pkg/rules/file_config.go
```diff
@@ -0,0 +1,49 @@
...
+	Hooks      map[string]any      `yaml:"hooks,omitempty"`
+	Operations map[string]any      `yaml:"operations,omitempty"`
+}
+
+// DecodeFile decodes a folder rule configuration. Reserved sections are
+// accepted while empty so configurations remain forward-compatible.
+func DecodeFile(content []byte) (FileConfig, error) {
+	var config FileConfig
+	decoder := yaml.NewDecoder(bytes.NewReader(content))
+	decoder.KnownFields(true)
```
aakarim, line 30: Two message nits from §2.1 ("unknown rule-level keys … are errors with a did-you-mean suggestion"). An unknown rule key currently surfaces as yaml's `line 3: field foo not found in type rules.RuleSpec` — a Go type name and no suggestion. Rewrite the `yaml.TypeError` into `rules.<name>: unknown key "foo" (did you mean "for"?)` using the same nearest-match helper `compile.go` uses for params. And a missing `version` reads `version: expected 1, got 0` — say `version: required (expected 1)` when it's absent.

pkg/openlore/authz.go
aakarim: Fixed in 43326c8. Recursive deletes now walk the target and require config.edit for every contained folder config. Added coverage for denied rw-only deletion, allowed editor deletion, and allowed deletion of a config-free folder.

pkg/openlore/vfs.go
aakarim: Fixed in 43326c8. rules.IsDirConfigPath now rejects any earlier .lore segment, and DirFS delegates to that single helper. Added both rules-level and DirFS regression tests for the nested path.

pkg/openlore/scoped_write_fs.go
aakarim: Fixed in 43326c8. mutationDeniedError now provides consistent os.ErrPermission results for direct config mutations, recursive config-tree deletion, scoped FS entry points, and Server.AdmitChangeSet. Added removal and AdmitChangeSet assertions.

pkg/openlore/authz.go
aakarim: Removed in 43326c8. Authorization now remains exclusively at the two real-identity entry points; rules middleware only validates config and handles post-commit invalidation.

pkg/openlore/rules_plugin.go
aakarim: Fixed in 43326c8 by removing the early invalidation. Cache invalidation now happens only in PostCommitMiddleware.

pkg/openlore/rules_plugin.go
aakarim: Fixed in 43326c8. The always-registered plugin now always reports rules 0.1.0; reportInfo and the empty-info logging special case are gone.

pkg/rules/engine.go
aakarim: Fixed in 43326c8. Options.Folders is now FolderLayerSource directly, with the assertions and unreachable error removed.

pkg/openlore/rules_plugin.go
aakarim: Fixed in 43326c8. Folder layers now expose LayersForDir; file and bundle callers pass the known directory explicitly, so admission no longer Stats each target.

pkg/openlore/rules_plugin.go
aakarim: Agreed. Kept the simpler exclusive miss path for Phase 3 as suggested; the cache still avoids IO after the first lookup.

pkg/rules/file_config.go
aakarim: Fixed in 43326c8. Rule-level unknown keys now use the shared nearest-match logic and omit Go type names; an absent version now reports version: required (expected 1). Added regression cases for both.

pkg/openlore/authz.go
```diff
@@ -444,6 +452,42 @@ func (s *Server) identityCanWrite(id Identity, action vfs.ChangeAction, p string
 	return false
 }
 
+func (s *Server) canEditDirConfigsInTree(id Identity, target string) bool {
+	allowed := true
+	err := vfs.WalkDir(s.merge, target, func(candidate string, _ *vfs.FileInfo, walkErr error) error {
```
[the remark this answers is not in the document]
aakarim: Fixed in 37e2f68. Normal session writes now preserve the real Identity through middleware and the write log, and authorization is re-run in the serialized pre-apply step immediately before commit. Attribution-only approved/deferred submissions fail closed if they mutate a folder config or recursively delete a currently governed tree.

pkg/openlore/rules_plugin.go
```diff
@@ -77,32 +189,50 @@ func ruleRoots(docset config.DocsetSpec) []string {
...
-			if _, err := engine.Effective(context.Background(), path.Join(root, "__rules_boot_check__.md")); err != nil {
+			if _, err := bootEngine.Effective(context.Background(), path.Join(root, "__rules_boot_check__.md")); err != nil {
 				return nil, fmt.Errorf("configuring rules: %w", err)
 			}
 		}
 	}
+	options := rules.Options{Registry: rules.DefaultRegistry(), Config: configLayers, Env: env, Logger: logger}
+	if fsys != nil {
+		options.Folders = newFolderRuleLayers(fsys, auth.Docsets)
+	}
```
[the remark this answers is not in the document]
aakarim: Fixed in 37e2f68. Folder-rule cache reads and keys now use canonical docset paths while each returned layer preserves the requested alias scope and origin. Alias and canonical invalidation therefore converge on the same cache entry. Added a regression test that first caches a missing config through the alias, writes the config through that alias, and validates governed content through it.

pkg/openlore/vfs.go
```diff
@@ -622,7 +644,19 @@ func atomicWrite(full string, content []byte) error {
 }
 
 func (d *DirFS) Stat(p string) (*vfs.FileInfo, error) {
-	if isTrashPath(vfs.CleanPath(p)) || hasReservedPath(p) || hasTraversal(p) {
+	clean := vfs.CleanPath(p)
+	if path.Base(clean) == ".lore" && !hasTraversal(p) {
```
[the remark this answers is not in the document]
aakarim: Fixed in 37e2f68. The special Stat and ReadDir handling for .lore is now gated by the canonical folder-config path shape, so nested package-state .lore directories stay hidden. Added regression coverage for both operations.

pkg/rules/engine.go
```diff
@@ -155,16 +272,34 @@ func (e *Engine) validateFile(ctx context.Context, fsys vfs.FileSystem, bundleRo
...
+			dir := path.Dir(path.Dir(file.AbsolutePath))
+			if err := e.CheckConfigFile(context.Background(), dir, file.Content); err != nil {
+				invalidDirs[dir] = true
+				out = append(out, validation.Diagnostic{Path: relative(bundle.Root, file.AbsolutePath), Line: 1, Column: 1, Severity: validation.SeverityError, Rule: "rules/config", Message: err.Error()})
+			}
+		}
+		var bundleRules []CompiledRule
+		if !invalidDirs[vfs.CleanPath(bundle.Root)] {
+			var err error
+			bundleRules, err = e.effectiveDir(context.Background(), bundle.Root)
```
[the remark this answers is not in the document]
aakarim: The implementation matches plan §3.4: a bundle rule is resolved at the caller-named bundle root. Thus /docs/sub/.lore/config.yaml applies when validating /docs/sub, not when validating parent bundle /docs. I added TestValidatorRunsBundleRuleWhenFolderIsNamedBundleRoot in 37e2f68 to lock down that behavior; no semantic change was needed.

PR #70, opened 2026-09-03
Implement folder rules phase 4

aakarim:
## Summary
- add symlink-safe per-package state storage under content-local `.lore/` directories
- implement append-only `max: initial` size baselines, growth enforcement, lifecycle handling, and tokenizer re-baselining
- make `mv` atomic through a batch change set and carry baseline state across moves
- add `lore size baseline` history/reset commands with authorization and audit logging
- publish the new size-rule parameters in generated stdlib docs

## Verification
- `nix develop --command go test ./...`
- `nix develop --command env GOOS=windows GOARCH=amd64 go build ./...`
- focused end-to-end server test for create baseline, allowed growth, and rejection

### DIFF
--- a/pkg/openlore/rules_plugin.go
+++ b/pkg/openlore/rules_plugin.go
@@ -13,10 +13,12 @@ import (
 
 	"github.com/aakarim/go-openlore/internal/config"
 	"github.com/aakarim/go-openlore/pkg/openlore/validation"
+	"github.com/aakarim/go-openlore/pkg/packagestate"
 	"github.com/aakarim/go-openlore/pkg/rules"
 	_ "github.com/aakarim/go-openlore/pkg/rules/link"
 	_ "github.com/aakarim/go-openlore/pkg/rules/okfrule"
 	_ "github.com/aakarim/go-openlore/pkg/rules/size"
+	"github.com/aakarim/go-openlore/pkg/rules/tokenizer"
 	"github.com/aakarim/go-openlore/pkg/vfs"
 )
 
@@ -213,6 +215,7 @@ func ruleRoots(docset config.DocsetSpec) []string {
 
 type rulesPlugin struct {
 	engine *rules.Engine
+	fs     vfs.FileSystem
 }
 
 func newRulesPlugin(auth *config.AuthConfig, defaults rules.Defaults, fsys vfs.FileSystem, logger *slog.Logger) (*rulesPlugin, error) {
@@ -222,7 +225,11 @@ func newRulesPlugin(auth *config.AuthConfig, defaults rules.Defaults, fsys vfs.F
 	}
 	sort.Slice(aliases, func(i, j int) bool { return len(aliases[i]) > len(aliases[j]) })
 	configLayers := configRuleLayers{global: auth.Rules, docsets: auth.Docsets}
-	env := func(string) rules.Env { return rules.Env{Defaults: defaults, Logger: logger, AliasRoots: aliases} }
+	mapper, _ := fsys.(packagestate.HostMapper)
+	env := func(member string) rules.Env {
+		pkg := strings.SplitN(member, "/", 2)[0]
+		return rules.Env{Defaults: defaults, State: packagestate.Open(mapper, pkg), Tokenizer: tokenizer.Estimator{}, Logger: logger, AliasRoots: aliases}
+	}
 	bootEngine := rules.New(rules.Options{Registry: rules.DefaultRegistry(), Config: configLayers, Env: env, Logger: logger})
 	// Compile every configured docset at boot so invalid members, parameters,
 	// and unification conflicts fail before the server begins accepting writes.
@@ -239,7 +246,7 @@ func newRulesPlugin(auth *config.AuthConfig, defaults rules.Defaults, fsys vfs.F
 		options.Folders = newFolderRuleLayers(fsys, auth.Docsets)
 	}
 	engine := rules.New(options)
-	return &rulesPlugin{engine: engine}, nil
+	return &rulesPlugin{engine: engine, fs: fsys}, nil
 }
 
 func (p *rulesPlugin) WriteMiddleware() []WriteMiddleware {
@@ -255,7 +262,7 @@ func (p *rulesPlugin) WriteMiddleware() []WriteMiddleware {
 					}
 					continue
 				}
-				if err := p.engine.AdmitLeaf(ctx, leaf, op.Attribution.String(), nil); err != nil {
+				if err := p.engine.AdmitLeaf(ctx, leaf, op.Attribution.String(), p.existing(leaf.Target)); err != nil {
 					return WriteResult{}, err
 				}
 			}
@@ -287,11 +294,24 @@ func (p *rulesPlugin) PreApply(attribution Attribution, changes vfs.ChangeSet) e
 			}
 			continue
 		}
-		if err := p.engine.AdmitLeaf(context.Background(), leaf, attribution.String(), nil); err != nil {
+		if err := p.engine.AdmitLeaf(context.Background(), leaf, attribution.String(), p.existing(leaf.Target)); err != nil {
 			return err
 		}
 	}
 	return nil
+}
+
+func (p *rulesPlugin) existing(target string) func() ([]byte, bool, error) {
+	return func() ([]byte, bool, error) {
+		if p.fs == nil {
+			return nil, false, nil
+		}
+		content, err := p.fs.ReadFile(target)
+		if errors.Is(err, os.ErrNotExist) {
+			return nil, false, nil
+		}
+		return content, err == nil, err
+	}
 }
 
 func (p *rulesPlugin) invalidateChanges(changes []vfs.Change) {
@@ -307,7 +327,37 @@ func (p *rulesPlugin) invalidateChanges(changes []vfs.Change) {
 func (p *rulesPlugin) PostCommitMiddleware() []PostCommitMiddleware {
 	return []PostCommitMiddleware{func(next PostCommitHandler) PostCommitHandler {
 		return func(ctx context.Context, info CommitInfo) error {
+			ctx = rules.WithActor(ctx, info.Attribution.String())
 			p.invalidateChanges(info.ChangeSet.Leaves())
+			leaves := info.ChangeSet.Leaves()
+			moved := map[int]bool{}
+			for i, remove := range leaves {
+				if remove.Action != vfs.ChangeActionRemoveAll || remove.RemoveAll == nil || remove.RemoveAll.Opts.Expected == nil {
+					continue
+				}
+				var sourceHash string
+				for _, op := range remove.RemoveAll.Opts.Expected.Ops {
+					if op.RelPath == "." && op.Kind == "file" {
+						sourceHash = op.Hash
+					}
+				}
+				for j, write := range leaves {
+					if write.Action == vfs.ChangeActionWrite && write.Write != nil && hashBytes(write.Write.Bytes) == sourceHash {
+						if err := p.engine.OnMove(ctx, remove.Target, write.Target); err != nil {
+							return err
+						}
+						moved[i], moved[j] = true, true
+						break
+					}
+				}
+			}
+			for i, leaf := range leaves {
+				if !moved[i] && (leaf.Action == vfs.ChangeActionRemove || leaf.Action == vfs.ChangeActionRemoveAll) {
+					if err := p.engine.OnRemove(ctx, leaf.Target); err != nil {
+						return err
+					}
+				}
+			}
 			return next(ctx, info)
 		}
 	}}
--- a/pkg/openlore/rules_plugin.go
+++ b/pkg/openlore/rules_plugin.go
@@ -301,7 +301,7 @@ func (p *rulesPlugin) PreApply(attribution Attribution, changes vfs.ChangeSet) e
 			}
 			continue
 		}
-		if err := p.engine.AdmitLeaf(context.Background(), leaf, attribution.String(), p.existing(leaf.Target)); err != nil {
+		if err := p.engine.PreApplyLeaf(context.Background(), leaf, attribution.String(), p.existing(leaf.Target)); err != nil {
 			return err
 		}
 	}
@@ -335,40 +335,36 @@ func (p *rulesPlugin) PostCommitMiddleware() []PostCommitMiddleware {
 	return []PostCommitMiddleware{func(next PostCommitHandler) PostCommitHandler {
 		return func(ctx context.Context, info CommitInfo) error {
 			p.invalidateChanges(info.ChangeSet.Leaves())
-			leaves := info.ChangeSet.Leaves()
-			movedRemoves := map[int]bool{}
-			for i, remove := range leaves {
-				if remove.Action != vfs.ChangeActionRemoveAll || remove.RemoveAll == nil || remove.RemoveAll.Opts.Expected == nil {
-					continue
-				}
-				var sourceHash string
-				for _, op := range remove.RemoveAll.Opts.Expected.Ops {
-					if op.RelPath == "." && op.Kind == "file" {
-						sourceHash = op.Hash
-					}
-				}
-				for _, write := range leaves {
-					if write.Action == vfs.ChangeActionWrite && write.Write != nil && hashBytes(write.Write.Bytes) == sourceHash {
-						// Destination admission may have recorded a create baseline. Move
-						// deliberately replaces it wholesale with the source history.
-						if err := p.engine.OnMove(ctx, remove.Target, write.Target); err != nil {
-							return err
-						}
-						movedRemoves[i] = true
-						break
-					}
-				}
-			}
-			for i, leaf := range leaves {
-				if !movedRemoves[i] && (leaf.Action == vfs.ChangeActionRemove || leaf.Action == vfs.ChangeActionRemoveAll) {
-					if err := p.engine.OnRemove(ctx, leaf.Target, info.Attribution.String()); err != nil {
-						return err
-					}
-				}
-			}
 			return next(ctx, info)
 		}
 	}}
+}
+
+// CommitState updates package-local state synchronously after content commits
+// and before the submitter receives success.
+func (p *rulesPlugin) CommitState(ctx context.Context, info CommitInfo) error {
+	leaves := info.ChangeSet.Leaves()
+	movedRemoves := map[int]bool{}
+	for _, move := range info.ChangeSet.Moves {
+		remove, write := leaves[move.From], leaves[move.To]
+		if err := p.engine.OnMove(ctx, remove.Target, write.Target); err != nil {
+			return err
+		}
+		movedRemoves[move.From] = true
+	}
+	for i, leaf := range leaves {
+		switch {
+		case leaf.Action == vfs.ChangeActionWrite && leaf.Write != nil && !rules.IsDirConfigPath(leaf.Target):
+			if err := p.engine.OnWrite(ctx, leaf.Target, leaf.Write.Bytes, info.Attribution.String()); err != nil {
+				return err
+			}
+		case (leaf.Action == vfs.ChangeActionRemove || leaf.Action == vfs.ChangeActionRemoveAll) && !movedRemoves[i]:
+			if err := p.engine.OnRemove(ctx, leaf.Target, info.Attribution.String()); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
 }
 func (p *rulesPlugin) Validators() []validation.Validator {
 	return []validation.Validator{p.engine.Validator()}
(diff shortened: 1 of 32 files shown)


pkg/openlore/rules_size.go
```diff
@@ -0,0 +1,150 @@
...
+		}
+		if !allowed {
+			return "", errors.New("permission denied")
+		}
+	}
+	return b.server.rules.baselineText(context.Background(), target)
+}
+func (b sessionSizeBackend) Reset(target, note string, a cmds.JobAttribution) (string, error) {
+	id := b.server.resolveSessionIdentity(b.identity)
+	if !scopeGrantsWrite(id.Scopes) || !b.server.identityHasDirConfigRole(id, target) || !b.server.identityCanWrite(id, vfs.ChangeActionWrite, target) {
```
aakarim: **Bug: reset is denied at the docset root.** `identityHasDirConfigRole(id, target)` expects a `<dir>/.lore/config.yaml` path — it does `path.Dir(path.Dir(target))` to find the governed directory. Passing the content path here means `/docs/a.md` → `/` → no docset → `false`. Verified: alice (`rw` + `config.edit`) gets `permission denied` for `lore size baseline reset /docs/a.md`, while `/docs/sub/n.md` happens to work because the double `Dir` lands on `/docs`. Pass `path.Join(path.Dir(target), ".lore", "config.yaml")`, or better, split the helper so it takes the governed directory and both call sites use it.

Same line: `identityCanWrite` already checks `scopeGrantsWrite`, so the leading `!scopeGrantsWrite(id.Scopes)` is redundant.

This also means the "no max: initial size rule applies" branch below is unreachable for root-level files — `lore size baseline reset /docs/x.txt` reports `permission denied` instead of naming the rules that apply (acceptance 12). Please add the shell tests for acceptance 10, 11 and 12 (happy path with `--note`, two `baseline` records with the first byte-for-byte unchanged, cap 22 after reset from 15 × 1.5, denied for `rw`-only and for `ro`+`config.edit`, no JSONL append and no audit event on denial).

pkg/rules/size/size.go
```diff
@@ -2,14 +2,144 @@ package size
...
+	for _, record := range history {
+		if record.Op == "remove" {
+			found = false
+		} else if record.Op == "baseline" {
+			current, found = record, true
+		}
+	}
+	return current, found, nil
+}
+func (s *baselineStore) Record(ctx context.Context, b Baseline) error {
```
aakarim, line 78: **Bug: zero timestamps.** `Record` fills `b.At` on its local copy, so the caller's `baseline` keeps a zero `At`. Two visible symptoms from the probes: the rule-added rejection reads `set 0001-01-01 on rule-added`, and the `rules.baseline.reset` audit event has `details.next.ts: "0001-01-01T00:00:00Z"` while `details.previous.ts` is real. Stamp `At` (and default `Op`) before calling `Record` — e.g. have `Record` take `*Baseline`, or return the stored record — and assert the date in the acceptance-3 test.

aakarim (commented):
Reviewed against `docs/rules-system.md` Phase 4 (§3.3, §3.7, §4.6–4.9, acceptance tests 1–14). On `1cc9231`: gofmt, `go vet ./...` and `go test ./...` are clean. I ran a throwaway server test over the acceptance items the PR doesn't cover: rule-added (Q28), remove/recreate, `mv` same-dir and cross-dir, `lore validate` read-only, reset happy/denied, history output. State carry through `mv`, remove/recreate and validate all behave as specified — nice.

Three things need fixing before merge, two of them found by the probes:

1. **`lore size baseline reset` is denied for everyone on files directly under a docset root** (wrong path handed to `identityHasDirConfigRole`). Acceptance test 10 fails as written.
2. **Baseline timestamps are zero in messages and audit events** (`Record` sets `At` on a copy).
3. **Rejected creates leave a phantom baseline** that later governs the real file. Adil chose option B (treat a baseline as absent when the file does not exist); details inline.

Test coverage: the PR adds acceptance 1, 2 (partial) and the `mv` batch shape; items 3, 4, 5, 7, 8, 10, 11, 12, 13 have no tests. Please add them — I've noted exact expectations inline where a probe showed something worth pinning.

pkg/rules/size/size.go
```diff
@@ -26,87 +156,177 @@ func Register(reg *rules.Registry) {
...
+	limit, provenance := c.max, fmt.Sprintf("max: %d", c.max)
+	if c.initial {
+		baseline, ok, err := c.store.Get(ctx, subject.Path)
+		if err != nil {
+			return nil, err
+		}
+		if c.metric == Tokens && ok && baseline.Tokenizer != c.tokenizer().Name() {
+			ok = false
+		}
+		if !ok {
```
aakarim: **Decision (Adil): option B — a baseline is absent when the file does not exist.** Today a `create` baseline is recorded during admission, before commit. If another rule then rejects the same write, the baseline stays. Verified: `size/lines {max: initial, growth: 1.5}` plus a fixed `size/kilobytes {max: 1}`; a 1000-line create is rejected by the KiB rule but leaves `{"reason":"create","lines":1000}`; the real 10-line create is then judged against 1000 × 1.5 and a 16-line rewrite passes when it must fail.

Fix: evaluate `subject.Existing()` first, and when the file does not exist ignore any baseline `Get` returned and fall through to the `create` path (record a fresh baseline from the proposed content). History keeps the phantom record — fine, it's append-only — but it never governs anything. Please add a test for exactly this sequence (rejected 1000-line create → 10-line create → 16-line write rejected with `baseline 10 lines`).

While here: `Existing()` is only consulted when no baseline is found, so its result can drive the decision above without an extra read on the common path.

pkg/rules/size/size.go
```diff
@@ -26,87 +156,177 @@ func Register(reg *rules.Registry) {
...
+				}
+			}
+			if subject.Mode == rules.ModeValidate {
+				return nil, nil
+			}
+			baseline, err = Measure(baseContent, c.tokenizer())
+			if err != nil {
+				return nil, err
+			}
+			baseline.Path, baseline.Reason, baseline.Actor = subject.Path, "create", subject.Actor
```
aakarim, line 287: When the token baseline is discarded because `baseline.Tokenizer != c.tokenizer().Name()` (line 235), the file *does* exist, so this path records the new baseline with `reason: "rule-added"`. That's misleading in `lore size baseline` history and in messages (`set … on rule-added`) — nothing was added. Use a distinct reason (`"tokenizer-changed"`, or carry the previous tokenizer name in the record) and document the value alongside `create`/`rule-added`/`reset` in §3.7.

Acceptance 13 also needs an injection point for a fake `tokenizer.Tokenizer`; today `rules_plugin.go:231` hard-codes `tokenizer.Estimator{}` and `Reset` in `rules_size.go:139` does too, so `reset` would record `estimate/v1` even after a real tokenizer is wired. Thread the tokenizer through the plugin (a server option is fine, test-only for now), have `Inspect` expose it, and use it in `Reset`.

pkg/rules/size/size.go
```diff
@@ -2,14 +2,144 @@ package size
...
+type Baseline struct {
+	Path      string    `json:"path,omitempty"`
+	At        time.Time `json:"ts"`
+	Op        string    `json:"op"`
+	Reason    string    `json:"reason,omitempty"`
+	Actor     string    `json:"actor,omitempty"`
+	Note      string    `json:"note,omitempty"`
+	Commit    string    `json:"commit,omitempty"`
+	Kilobytes int       `json:"kilobytes,omitempty"`
+	Lines     int       `json:"lines,omitempty"`
```
aakarim: `omitempty` on `kilobytes`/`lines`/`tokens` drops the field for a 0-line or empty file, so a reader can't tell "measured 0" from "not measured". §3.7 shows all three always present on `baseline` records. Keep `omitempty` for `remove` records by splitting the two record shapes, or just drop it on the three metrics (a `remove` record with `lines: 0` is unambiguous given `op`).

pkg/packagestate/store.go
```diff
@@ -0,0 +1,231 @@
...
+	to, ok := target.(*dirStore)
+	if !ok {
+		return os.ErrPermission
+	}
+	var records [][]byte
+	for record, err := range d.Records(key) {
+		if err != nil {
+			return err
+		}
+		records = append(records, record)
```
aakarim: `Move` copies records verbatim, so after `mv /docs/a.md /docs/b.md` the carried record still says `"path":"/docs/a.md"` inside `b.md.jsonl` (verified; same for the cross-directory move into `sub/c.md.jsonl`). The store is generic and shouldn't know about `path`, so either have `size.BaselineStore.Move` rewrite `Path` on the way through (it decodes anyway), or drop `path` from the record since the key already identifies the file. Please also add the acceptance-5 server test with *distinguishable* baselines (create at 10, grow to 15, `mv`, then assert the destination cap is still 15 not 22) — with equal sizes the test can't tell a carried baseline from a fresh `create`, which is exactly the failure mode admission-then-post-commit ordering could produce.

pkg/openlore/rules_plugin.go
```diff
@@ -307,7 +327,37 @@ func (p *rulesPlugin) invalidateChanges(changes []vfs.Change) {
...
+					continue
+				}
+				var sourceHash string
+				for _, op := range remove.RemoveAll.Opts.Expected.Ops {
+					if op.RelPath == "." && op.Kind == "file" {
+						sourceHash = op.Hash
+					}
+				}
+				for j, write := range leaves {
+					if write.Action == vfs.ChangeActionWrite && write.Write != nil && hashBytes(write.Write.Bytes) == sourceHash {
```
aakarim: Ordering note worth a comment here: admission of the destination `Write` leaf runs `Evaluate` first and records a fresh `create` baseline for the destination, then this post-commit `OnMove` → `Replace` overwrites it with the carried records. It works because `Replace` is wholesale, but it's implicit — a future `Append`-based `Move` would silently break carry. Two lines of comment, plus the distinguishable-size test mentioned on `store.go`, would pin it.

Also: `moved[j] = true` is set but never read for writes (only removes are consulted in the second loop).

pkg/openlore/rules_plugin.go
```diff
@@ -222,7 +225,11 @@ func newRulesPlugin(auth *config.AuthConfig, defaults rules.Defaults, fsys vfs.F
 	}
 	sort.Slice(aliases, func(i, j int) bool { return len(aliases[i]) > len(aliases[j]) })
 	configLayers := configRuleLayers{global: auth.Rules, docsets: auth.Docsets}
-	env := func(string) rules.Env { return rules.Env{Defaults: defaults, Logger: logger, AliasRoots: aliases} }
+	mapper, _ := fsys.(packagestate.HostMapper)
+	env := func(member string) rules.Env {
+		pkg := strings.SplitN(member, "/", 2)[0]
+		return rules.Env{Defaults: defaults, State: packagestate.Open(mapper, pkg), Tokenizer: tokenizer.Estimator{}, Logger: logger, AliasRoots: aliases}
```
aakarim: Nit: `pkg` is derived by splitting the member path on the first `/`. That's the Go rule for stdlib, but a VCS-namespaced member (`github.com/acme/rules/foo`) would get `github.com` as its state directory. Not reachable this release (packages are compiled-in), but a `rules.PackageOf(member)` helper next to `IsStdlib` would make the intent explicit and give the future resolver one place to change.

pkg/openlore/rules_size.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	history, err := store.History(ctx, target)
+	if err != nil {
+		return "", err
+	}
+	if len(history) == 0 {
+		return "no baseline recorded\n", nil
+	}
+	var b strings.Builder
+	fmt.Fprintln(&b, target)
```
aakarim: §3.7 specifies the header as `<path>  (size/lines via <rule> @ <origin>, growth 1.25)` so a reader can see *which* rule and layer produce the cap. `states[i].rule` already carries `Spec`/origin — print it. Also `Reset` (line 146) reports only `states[0]`; when both `size/lines` and `size/tokens` are `max: initial` on the same file the one record resets both, so print every applicable cap like `baselineText` does.

pkg/rules/engine.go
```diff
@@ -364,6 +394,8 @@ func matchesAtScope(rule CompiledRule, target string) bool {
 	return ok && Matches(rule.Spec, rel)
 }
 
+func Applies(rule CompiledRule, target string) bool { return matchesAtScope(rule, target) }
```
aakarim: Nit: `Applies` is an exported alias for `matchesAtScope`; just export `matchesAtScope` under the new name rather than keeping two.

pkg/rules/types.go
```diff
@@ -98,6 +102,16 @@ const (
 	ModeValidate
 )
 
+type actorContextKey struct{}
+
+func WithActor(ctx context.Context, actor string) context.Context {
```
aakarim: The actor now arrives two ways: `Subject.Actor` for `Evaluate`, and a context value for `OnRemove`. Since only removes need this and the engine's `OnRemove`/`OnMove` are new in this PR, I'd rather add `actor string` to `Engine.OnRemove`/`Check.OnRemove` than introduce a context key — the plan's `Check` signature is ours to change, and a typed parameter can't be silently missing.

pkg/openlore/rules_plugin_test.go
```diff
@@ -138,6 +138,34 @@ func admitServer(server *Server, target, content string) error {
 	return err
 }
 
+func TestInitialSizeBaselinePersistsAndRejectsGrowth(t *testing.T) {
```
aakarim, line 142: Coverage: this is acceptance 2 (good — it checks actor and `reason: create`). Missing from the PR are acceptance 3 (rule-added: 20-line file, rule added, 40-line write rejected `40 > 30` *and* a `rule-added` record with `lines: 20`, then 30 succeeds — my probe confirms the behaviour, so this is pinning), 4 (remove appends `remove`; recreate at 100 → cap 150), 5 (`mv`, see `store.go` comment), 7 (`lore validate /docs` creates no `.lore/` — confirmed), 8 (embedded read-only content, no `.lore` created), 10–12 (reset/history shell tests), 13 (tokenizer change). Since post-commit runs after `Submit` returns, tests that read the host `.lore/size/*.jsonl` right after a `mv` or `rm` need to wait for the post-commit chain (a second `Submit` on the same log is enough, since the applier is single-threaded) — I hit a half-written `b.md.jsonl.tmp` when reading immediately.

pkg/openlore/rules_size.go
aakarim: Fixed in e69acfe: reset now checks the governing directory config path, removes the redundant scope check, and has happy/denied/ungoverned shell coverage.

pkg/rules/size/size.go
aakarim: Fixed in e69acfe: admission and reset records are timestamped before storage, with rule-added output and reset audit coverage guarding against zero dates.

pkg/rules/size/size.go
aakarim: Fixed in e69acfe using option B: admission checks current existence even when history exists and ignores phantom baselines for absent files. Added the exact rejected-create regression sequence.

pkg/rules/size/size.go
aakarim: Fixed in e69acfe: tokenizer changes record tokenizer-changed, tokenizer injection is threaded through server/plugin/check/reset, and non-token metrics ignore tokenizer-only rebaseline records. Added regression coverage.

pkg/rules/size/size.go
aakarim: Fixed in e69acfe: all three measured fields are always serialized, including zero values.

pkg/packagestate/store.go
aakarim: Fixed in e69acfe: size state rewrites Path while moving records. The cross-directory server test uses a 10-line baseline, grows to 15, moves, and verifies 16 is still rejected against 10.

pkg/openlore/rules_plugin.go
aakarim: Addressed in e69acfe: documented the replacement ordering inline, removed the unused write index tracking, and added the distinguishable cross-directory move test.

pkg/openlore/rules_plugin.go
aakarim: Addressed in e69acfe with rules.PackageOf plus stdlib and VCS-path tests.

pkg/openlore/rules_size.go
aakarim: Fixed in e69acfe: history headers include member/rule/origin/growth, and reset output reports every applicable cap.

pkg/rules/engine.go
aakarim: Addressed in e69acfe: Applies is now the implementation rather than an alias.

pkg/rules/types.go
aakarim: Addressed in e69acfe: remove actor is now an explicit typed parameter through Engine and Check; the context key was removed.

pkg/openlore/rules_plugin_test.go
aakarim: Added coverage in e69acfe for rule-added, phantom rejected creates, remove/recreate, distinguishable cross-directory moves, validate read-only behavior, embedded content, reset happy/denied/ungoverned/history, and tokenizer changes.

pkg/openlore/rules_plugin.go
```diff
@@ -308,6 +335,37 @@ func (p *rulesPlugin) PostCommitMiddleware() []PostCommitMiddleware {
...
+				for _, op := range remove.RemoveAll.Opts.Expected.Ops {
+					if op.RelPath == "." && op.Kind == "file" {
+						sourceHash = op.Hash
+					}
+				}
+				for _, write := range leaves {
+					if write.Action == vfs.ChangeActionWrite && write.Write != nil && hashBytes(write.Write.Bytes) == sourceHash {
+						// Destination admission may have recorded a create baseline. Move
+						// deliberately replaces it wholesale with the source history.
+						if err := p.engine.OnMove(ctx, remove.Target, write.Target); err != nil {
```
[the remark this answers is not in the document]
aakarim: Fixed in 67649e1. Moves now carry explicit validated source/destination leaf indices in the serialized ChangeSet. The metadata requires a snapshot-guarded remove whose file hash matches the paired write, so identical-content operations cannot be associated ambiguously.

pkg/packagestate/store.go
```diff
@@ -0,0 +1,231 @@
...
+
+type store struct {
+	mapper HostMapper
+	pkg    string
+}
+
+// Open scopes a store to one package. Uppercase letters use Go module-cache
+// escaping (Github.com/X becomes !github.com/!x).
+func Open(mapper HostMapper, pkg string) Store {
+	return &store{mapper: mapper, pkg: escapePath(pkg)}
```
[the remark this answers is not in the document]
aakarim: Fixed in 67649e1. Package identifiers are validated before any host path is constructed; empty, dot/dot-dot, rooted, unclean, backslash, NUL, and drive/ADS-style colon paths are rejected. Added traversal/path tests.

pkg/rules/size/size.go
```diff
@@ -2,14 +2,167 @@ package size
...
+		}
+		records = append(records, rewritten)
+	}
+	if len(records) == 0 {
+		return nil
+	}
+	if err := dst.Replace(stateKey(to), records); err != nil {
+		return err
+	}
+	return src.Remove(stateKey(from))
```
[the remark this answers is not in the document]
aakarim: Fixed in 67649e1. Package stores now share a package-level mutex across Open calls and all append/read/replace/remove operations. Size history migration uses RewriteMove, which performs read, path rewrite, destination replacement, and source removal under that one lock.

pkg/openlore/rules_size.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+	current, ok, err := store.Get(ctx, target)
+	if err != nil {
+		return "", err
+	}
+	if ok && len(states) != 0 {
+		fmt.Fprint(&b, "current cap:")
+		for _, state := range states {
+			value, unit := baselineMetric(current, state.metric)
+			fmt.Fprintf(&b, " %d %s", int(math.Floor(float64(value)*state.growth)), unit)
```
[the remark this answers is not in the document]
aakarim: Fixed in 67649e1. The size package now exposes Current, which applies the same metric and tokenizer-aware baseline selection as enforcement. History and reset resolve the prior/current baseline independently for every compiled size rule.

aakarim:
Addressed the additional review findings in `67649e1`:

- Package-state lifecycle work now runs synchronously on the serialized write applier after content commit and before success is returned; state errors are surfaced. Baseline reset uses the same applier boundary, so its read+append cannot race writes.
- Admission is side-effect-free. Rule-added/tokenizer migration records occur only during serialized pre-apply, while create baselines are recorded exactly once after a successful commit.
- Reset audit remains best-effort and is logged rather than turning an applied reset into a reported failure.
- Non-finite growth values are rejected.
- Added focused coverage for lifecycle timing, move metadata validation, unsafe package paths, and non-finite growth.

Verification: `go test ./...`, `go vet ./...`, and `GOOS=windows GOARCH=amd64 go build ./...` all pass.

Issue #65, opened 2026-09-02
OpenLore: agent-discoverability audit — no /llms.txt, .well-known/mcp.json, or docs mirrors (draft included)


ux-xd:
Hi — I maintain a handful of published MCP servers and spent this week auditing how well community servers' **websites** serve the agents that discover them (the directories list your server; the site is where an agent actually evaluates it). OpenLore scored high on product, and there are a couple of cheap wins:

**Five-point agent-readiness audit of https://openlore.sh** (run 2026-09-02, re-verified after filing):

| Check | Result |
|---|---|
| `/llms.txt` | 404 |
| `/.well-known/mcp.json` | 404 |
| MCP HTTP endpoint | POST /mcp answers 401 — something listens there; if it is your MCP endpoint it is gated |
| docs `.md` mirrors / `describedby` links | none |
| npm keyword hygiene | no package for this repo; unrelated 'openlore' package by clay-good exists (disambiguated, not counted) |

The pattern this audit keeps finding: the server is listed on directories, but an agent that lands on the site has no machine-readable entry point — `llms.txt` (llmstxt.org v2) plus a `.well-known/mcp.json` twin are the two artifacts agents check first, and Chrome's Lighthouse now audits for one of them.

**Draft `/llms.txt`** — take it as-is or treat it as a starting point:

```markdown
# OpenLore
> A minimal, extensible, agent-native knowledge base that keeps shared context current and inspectable

- Repo: https://github.com/aakarim/OpenLore
```

If it's useful I'm happy to open a PR with the file(s) instead. And if stdio-only is a deliberate choice, the endpoint row is an observation, not a complaint — close with no hard feelings.

*Correction (2026-09-02, after filing): note added: an unrelated npm package 'openlore' (clay-good/openlore) exists — not this project; not counted for or against.*

*Filed by an automated readiness audit, hand-verified claims. Method + all 10 probe cards: https://gist.github.com/ux-xd/f6408272f9cd643ddd1d698363d2412a*

aakarim:
Thanks for the PR. 

This repo is for users to create their own servers, so it's up to them whether they want to include llms.txt. 

The public site has diverged from this repo and will be removed shortly. We won't have a public repo for that and we'll handle AEO/SEO separately.

PR #71, opened 2026-09-04
Add /llms.txt

ux-xd:
Following up on #65 — this adds the `/llms.txt` drafted in that issue to `assets/web/`, which is embedded with the rest of the site assets and served from the root by the file server, so it answers at https://openlore.sh/llms.txt on the next deploy. One new file, nothing else changed. Happy to tweak the wording.


aakarim:
Closed per https://github.com/aakarim/OpenLore/issues/65#issuecomment-5547395697

PR #90, opened 2026-09-13
Add phase 1 usage metrics and analytics

aakarim:
## Summary
- add the experimental analytics event log, pipeline, aggregation registry, materialized store, and Prometheus exporter
- instrument sessions, authentication, commands, writes, exact content-change scalars, and correlation IDs
- expose analytics through shell commands, authenticated HTTP APIs, a plain HTML dashboard, file facts, and browser facts
- add graceful signal-driven shutdown and enable the analytics gate for the Fly deployment

## Verification
- `go test ./...`
- `go vet ./...`
- `go test -race ./internal/analytics ./pkg/openlore ./pkg/shell/...`
- rendered dashboard inspected with observed command, content facts, health values, and planned aggregation states


internal/analytics/pipeline.go
```diff
@@ -0,0 +1,257 @@
...
+			checkpointBytes, _ := os.ReadFile(p.checkpoint)
+			checkpointID := strings.TrimSpace(string(checkpointBytes))
+			pastCheckpoint := checkpointID == ""
+			_ = p.log.Scan(ctx, EventFilter{}, func(e Event) error {
+				if !pastCheckpoint {
+					p.consumePersisted(ctx, e)
+					pastCheckpoint = e.ID == checkpointID
+					return nil
+				}
+				p.handle(ctx, e)
```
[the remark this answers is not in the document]
aakarim: Fixed. Startup now detects when the saved event ID is absent from retained segments and processes the full retained tail; regression coverage was added.

internal/analytics/pipeline.go
```diff
@@ -0,0 +1,257 @@
...
+	select {
+	case <-p.done:
+		return nil
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
+func (p *Pipeline) Replay(ctx context.Context, from time.Time) error {
+	p.seen = sync.Map{}
+	return p.log.Scan(ctx, EventFilter{From: from}, func(e Event) error { p.handle(ctx, e); return nil })
```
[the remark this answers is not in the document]
aakarim: Fixed. Replay and live handling now share one processing lock; replay also resets processor/consumer state and rewinds the history cursor for deterministic rebuilding.

internal/analytics/storage.go
```diff
@@ -0,0 +1,368 @@
...
+	if opts.Compress != "zstd" && opts.Compress != "none" {
+		return nil, fmt.Errorf("unsupported analytics compression %q", opts.Compress)
+	}
+	if err := os.MkdirAll(dir, 0o700); err != nil {
+		return nil, err
+	}
+	return &fileEventLog{dir: dir, opts: opts}, nil
+}
+func (l *fileEventLog) activePath(t time.Time) string {
+	return filepath.Join(l.dir, "events-"+t.UTC().Format("2006-01-02")+".jsonl")
```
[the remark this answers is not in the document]
aakarim: Fixed. Segment boundaries now use the configured rotation duration, while the default 24-hour rotation retains the documented daily filename format.

internal/config/config.go
```diff
@@ -549,6 +602,69 @@ type fileConfig struct {
...
+	} `yaml:"export"`
+}
+
+func parseAnalyticsDuration(value, name string, target *time.Duration) error {
+	if value == "" {
+		return nil
+	}
+	d, err := time.ParseDuration(value)
+	if err != nil || d < 0 {
+		return fmt.Errorf("invalid analytics %s %q", name, value)
```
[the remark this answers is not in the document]
aakarim: Fixed. Analytics durations now accept non-negative day values such as `90d`, with config coverage for the checked-in example.

pkg/openlore/analytics_scalars.go
```diff
@@ -0,0 +1,111 @@
...
+func (p *ScalarProcessor) Process(ctx context.Context, e analytics.Event) []analytics.Event {
+	if e.Type != "doc.write" {
+		return nil
+	}
+	target, _ := e.Fields["commit_id"].(string)
+	var out []analytics.Event
+	for {
+		record, ok, err := p.history.Next(ctx)
+		if err != nil || !ok {
+			return out
```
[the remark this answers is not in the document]
aakarim: Fixed. The atomic pipeline checkpoint now includes each processor cursor, including the history byte offset and last commit ID, and restores it before startup processing.

pkg/openlore/analytics_scalars.go
```diff
@@ -0,0 +1,111 @@
...
+			for k := range keys {
+				delta[k] = after[k] - before[k]
+			}
+			parent := ""
+			invocation := record.ID
+			if record.ID == target {
+				parent = e.ID
+				invocation = e.InvocationID
+			}
+			out = append(out, analytics.Event{ID: analytics.NewID(), Time: time.Now().UTC(), Type: "doc.scalars", Principal: record.Attribution.Principal, Actor: record.Attribution.Actor, InvocationID: invocation, ParentID: parent, Fields: map[string]any{"path": leaf.Target, "docset": docsetFromPath(leaf.Target), "action": action, "writer": string(writer), "commit_id": record.ID, "content_hash": contentHash, "before": before, "after": after, "delta": delta, "tokenizer": "approx", "first_seen": firstSeen}})
```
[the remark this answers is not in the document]
aakarim: Fixed. History commits and `(commit_id, path)` outputs are processed idempotently, so subsequent leaf events for the same commit cannot advance or misattribute the cursor.

pkg/shell/cmds/ls.go
```diff
@@ -40,7 +50,25 @@ func CmdLs(ctx CmdContext, args []string, w io.Writer, errW io.Writer, stdin io.
...
 			if longFormat {
-				PrintLong(w, f)
+				if statsFlag && analytics != nil {
+					facts, factErr := analytics.Facts().Stat(context.Background(), p)
+					if factErr != nil {
+						fmt.Fprintf(errW, "ls: %s: %s\n", target, factErr)
+						exitCode = 1
+						continue
+					}
+					fmt.Fprintf(w, "%s %8.0f %8.0f %s\n", f.Mode(), facts.Scalars["lines"], facts.Scalars["tokens"], f.Name())
```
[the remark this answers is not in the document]
aakarim: `ls` analytics facts are now created from the command session’s `ctx.FS()`, with a regression test proving the global filesystem is not consulted.

pkg/shell/cmds/stat.go
```diff
@@ -17,6 +25,15 @@ func CmdStat(ctx CmdContext, args []string, w io.Writer, errW io.Writer, stdin i
 			fmt.Fprintf(errW, "stat: %s: %s\n", a, err)
 			return 1
 		}
+		if jsonFlag && analytics != nil {
+			facts, factErr := analytics.Facts().Stat(context.Background(), p)
+			if factErr != nil {
+				fmt.Fprintf(errW, "stat: %s: %s\n", a, factErr)
+				return 1
+			}
+			_ = json.NewEncoder(w).Encode(facts)
```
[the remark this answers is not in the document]
aakarim: `stat` analytics facts are now created from the command session’s `ctx.FS()`. HTTP current-content facts were also narrowed to the authenticated session filesystem.

internal/analytics/aggregations.go
```diff
@@ -0,0 +1,296 @@
...
+	}
+	return Table{Columns: []string{"bucket", "sessions", "new_principals", "commands", "unknown_commands"}, Rows: limitRows(rows, p), Total: len(rows)}, nil
+}
+func treeSize(ctx context.Context, _ EventSource, facts ContentFacts, p Params) (Table, error) {
+	root := p.Extra["path"]
+	if root == "" {
+		root = "/"
+	}
+	var rows [][]any
+	err := facts.Walk(ctx, root, WalkOptions{}, func(d DocScalars) error {
```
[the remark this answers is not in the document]
aakarim: Fixed. `tree-size` parses its `depth` parameter and passes it through to `WalkOptions.Depth`; focused coverage verifies the requested depth.

internal/analytics/aggregator.go
```diff
@@ -0,0 +1,111 @@
...
+	switch e.Type {
+	case "command.exec":
+		cmd := fieldString(e, "command")
+		exitClass := "success"
+		if fieldFloat(e, "exit_code") != 0 {
+			exitClass = "error"
+		}
+		key := cmd + "\x00" + e.Transport + "\x00" + exitClass
+		a.commands[key]++
+		a.durations[cmd] = append(a.durations[cmd], fieldFloat(e, "duration_ms")/1000)
```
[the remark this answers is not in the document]
aakarim: Fixed. Command durations now retain only a sum and count per command, keeping both memory and scrape work bounded.

internal/analytics/pipeline.go
```diff
@@ -0,0 +1,257 @@
...
+				return nil
+			})
+			ticker := time.NewTicker(time.Second)
+			defer ticker.Stop()
+			for {
+				select {
+				case e := <-p.handoff:
+					p.handle(ctx, e)
+				case <-ticker.C:
+					_ = p.log.Scan(ctx, EventFilter{}, func(e Event) error { p.handle(ctx, e); return nil })
```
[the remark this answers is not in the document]
aakarim: Fixed. The file-backed pipeline maintains per-segment byte offsets and decodes only newly appended data on periodic scans.

pkg/openlore/analytics_plugin.go
```diff
@@ -0,0 +1,226 @@
...
+				action = "create"
+				for _, record := range info.Leaves {
+					if vfs.CleanPath(record.Target) == vfs.CleanPath(leaf.Target) && record.BeforeHash != "" {
+						action = "update"
+						break
+					}
+				}
+			} else {
+				action = "delete"
+			}
```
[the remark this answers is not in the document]
aakarim: Fixed. `doc.write` and history scalar processing now include only write/remove/remove_all content leaves; explicit namespace and metadata actions are ignored.

pkg/openlore/server.go
```diff
@@ -1079,7 +1132,30 @@ func (s *Server) buildSessionShell(id Identity) *shell.Shell {
...
+			sh.SetAnalytics(s.analytics)
+		}
+		sh.SetCommandObserver(func(ce shell.CommandExecution) {
+			s.metrics.TotalCommands.Add(1)
+			e := s.analyticsEvent(id, "command.exec", map[string]any{"command": ce.Command, "argc": ce.Argc, "exit_code": ce.ExitCode, "duration_ms": ce.Duration.Milliseconds(), "bytes_out": ce.BytesOut, "pipeline_position": ce.PipelinePosition})
+			e.ID = ce.EventID
+			e.InvocationID = ce.InvocationID
+			s.analytics.Record(context.Background(), e)
 		})
 	}
```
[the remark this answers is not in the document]
aakarim: Fixed. Every session shell now installs the legacy command counter observer; analytics event emission alone remains conditional.

pkg/openlore/server.go
```diff
@@ -1515,7 +1596,11 @@ func (s *Server) ListenAndServe() error {
 	s.srv = srv
 
 	if s.config.MetricsPort > 0 {
-		metrics.StartServer(s.config.MetricsPort, s.metrics, s.logger)
+		if s.analytics != nil && s.config.Analytics.Export.Prometheus {
+			s.metricsSrv = metrics.StartHandlerServer(s.config.MetricsPort, s.analytics.Aggregator(), s.logger)
```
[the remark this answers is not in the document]
aakarim: Fixed without breaking the existing contract: `/metrics` continues serving the established JSON metrics and analytics Prometheus exposition is composed at `/metrics/prometheus`.

pkg/openlore/server.go
```diff
@@ -1719,15 +1824,30 @@ func (s *Server) Shutdown(ctx context.Context) error {
...
 	if s.writeLog != nil {
 		if err := s.writeLog.Close(ctx); err != nil {
 			s.logger.Warn("shutdown: write log did not drain before deadline", "err", err)
 		}
 	}
-	if s.srv == nil {
-		return nil
+	if s.analytics != nil {
+		if err := s.analytics.Close(ctx); err != nil {
+			s.logger.Warn("shutdown: analytics did not drain before deadline", "err", err)
```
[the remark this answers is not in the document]
aakarim: Fixed. Analytics close now gets a dedicated context bounded by `analytics.shutdown_timeout` (with the configured default fallback).

pkg/openlore/writelog.go
```diff
@@ -117,17 +123,28 @@ func (l *writeLog) run() {
 			err = pre(e.identity, e.attribution, e.cs)
 		}
 		if err == nil {
+			leaves, captureErr := capturePreImages(context.Background(), l.substrate, l.blobs, e.cs, l.blobsEnabled)
```
aakarim: **Bug (blocks decision 25/26):** `leaves, captureErr := …` declares a *new* `leaves` inside the `if`, shadowing the `var leaves []LeafRecord` on line 110. The `_ = leaves` on line 130 is what silences the compiler. The outer `leaves` stays nil, so:

- `CommitInfo.Leaves` is always empty → `observeWrites` reports every write as `create`.
- `CommitRecord.Leaves` in `history/commits.jsonl` is always empty → `ScalarProcessor.processRecord` iterates nothing → **no `doc.scalars` event is ever emitted in a running server**, and `size-over-time`, `write-ratio`, `openlore_doc_write_scalar_total` are all permanently empty.
- The blob store is still populated (the side effect happens inside `capturePreImages`), so `before_hash` blobs exist with nothing pointing at them.

I reproduced it with a throwaway test on this branch that submits an overwrite of `/doc.md` through `newWriteLog(...).Submit` with `SetCommitJournal(path, blobs, true)` and reads the journal back:

```
journal leaves=0 record={"id":"01a09fdd…","time":…,"attribution":{"principal":"adil"},"change_set":{…},"hash":"f395…"}
CommitInfo.Leaves=0
expected one leaf with before_hash, got []openlore.LeafRecord(nil)
```

Fix is `var captureErr error; leaves, captureErr = capturePreImages(...)` and drop `_ = leaves`. Please also add the applier-level test the plan lists ("pre-image capture A→B→delete" through the write log, asserting `Leaves[0].BeforeHash`, `AfterHash`, and a `doc.scalars` event with a non-nil `before`) — the existing tests construct `CommitRecord.Leaves` by hand, which is why this slipped through.

pkg/openlore/history_blobs.go
```diff
@@ -0,0 +1,300 @@
...
+	"path"
+	"path/filepath"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/aakarim/go-openlore/pkg/vfs"
+)
+
+type CommitRecord struct {
```
aakarim, line 22: **Design question (decision 25 premise has shifted):** the plan said change scalars come from "the existing commit history (`history/commits.jsonl`, which already records every committed ChangeSet with its bytes)". On `release/v0.7.0` that file is now *legacy* — `history.go` migrates it into `events.jsonl` + per-file shards and `HistoryRecord` explicitly says payloads "belong in the filesystem or a separate content-addressed store, not in the query index". This PR re-creates `commits.jsonl` as a second journal, so the applier now writes two histories per commit (`history.Record` and `appendCommitRecord`), and the new one stores every write's full bytes forever (`ChangeSet.Write.Bytes`), which is exactly what the history split removed.

Since the blob store already exists, the plan-consistent shape is probably: store the *after*-image in `BlobStore` too (same content-addressed store, and it's what `restore --to <commit_id>` needs anyway), extend `HistoryRecord` (or the `events.jsonl` line) with `commit_id`, `before_hash`, `after_hash`, `before_size`, `after_size`, `before_unknown`, and have `HistoryCursor` read `events.jsonl` — `ScalarProcessor` then computes `after` from `blobs.Get(after_hash)` instead of from bytes embedded in the journal. That keeps one history, no payloads in the index, and no unbounded `commits.jsonl`. If you'd rather keep a separate "change log" file, the plan should be updated to say so and to name its retention story (Phase 3 lists history rotation, but this file grows with content size, not event count).

pkg/openlore/server.go
```diff
@@ -321,6 +340,19 @@ func newServerWithRoot(rootDir string, rootFS, lowerFS vfs.FileSystem, opts ...c
 			logger.Info("legacy history migrated")
 		}
 		s.writeLog = newWriteLog(s.merge, s.postCommitChain(), logger, 0)
+		if s.analytics != nil {
+			blobs, blobErr := OpenBlobStore(filepath.Join(dataDir, "history", "objects"))
+			if blobErr != nil {
+				return nil, fmt.Errorf("opening history blob store: %w", blobErr)
+			}
+			commitPath := filepath.Join(dataDir, "history", "commits.jsonl")
+			cursor, cursorErr := OpenHistoryCursor(commitPath, HistoryPosition{})
```
aakarim, line 350: **Coupling (decision 26):** the blob store, pre-image capture and the `Leaves` journal are only wired when `s.analytics != nil`, i.e. behind `experimental: [analytics]` *and* `analytics.enabled`. The plan makes the history/blob store part of the commit path ("Readonly servers have no history recorder and therefore no blobs" is the only exclusion) and says it is "the foundation for versioning and rollback". As written, a server that runs with analytics off for a while has no pre-images for those commits, so a later `restore` can't reach them, and turning analytics on later can't backfill `before` scalars. Suggest gating on `!cfg.Readonly` (+ `analytics.history.blobs`) rather than on the analytics service, and only the `ScalarProcessor` on analytics.

pkg/openlore/history_blobs.go
```diff
@@ -0,0 +1,300 @@
...
+				if enabled && blobs != nil {
+					if e = blobs.Put(ctx, record.BeforeHash, strings.NewReader(string(b))); e != nil {
+						record.BeforeUnknown = true
+					}
+				}
+			}
+		} else if err != nil && !errors.Is(err, os.ErrNotExist) {
+			record.BeforeUnknown = true
+		}
+		if !enabled && record.BeforeHash != "" {
```
aakarim: **Minor (decision 26 / owner requirement "the commit log should contain the previous hash of the file"):** with `analytics.history.blobs: false` this clears `BeforeHash` and marks `BeforeUnknown`. The plan says `blobs: false` "records leaves without pre-images" — the hash itself is already computed here and is cheap; only the blob write is the cost being switched off. Keeping `before_hash` (and `before_size`) and just skipping `blobs.Put` preserves the create/update distinction and `first_seen`, and lets `ScalarProcessor` fall back to bytes-only deltas from `before_size`.

internal/analytics/pipeline.go
```diff
@@ -0,0 +1,404 @@
...
+	}
+	cancel()
+	select {
+	case <-s.done:
+		return nil
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
+func (s *Shipper) ShipNow(ctx context.Context) error {
```
aakarim, line 478: **Deviation (decision 13, §Storage 2, Phase 1 wiring):** the Shipper does not ship. `ShipNow` only seals the previous day's segment, `remote` is never read, there is no shipped-offset checkpoint, and `Lag()` hard-codes 0. The plan is explicit that Phase 1 builds the loop and exercises it with a no-op `remote: none` driver so that the S3 driver is a drop-in: "byte-range deltas of the active segment as immutable objects, then the whole sealed object replaces the day's deltas. A checkpoint records shipped offsets so restart resumes", and `openlore_analytics_ship_lag_bytes` is one of the health series the deploy smoke test reads. Right now adding S3 later means writing the shipping algorithm, its checkpointing, and its failure/retry semantics from scratch, which is the drastic-later-change the phasing was meant to avoid. Sealing/retention also arguably belongs to the EventLog on a timer, not to the Shipper, so that `pipeline.enabled: false` plus `remote: none` still rotates.

internal/analytics/pipeline.go
```diff
@@ -0,0 +1,404 @@
...
+func (r *Refresher) Refresh(ctx context.Context, names ...string) error {
+	wanted := map[string]bool{}
+	for _, n := range names {
+		wanted[n] = true
+	}
+	for _, a := range r.registry.List() {
+		if len(wanted) > 0 && !wanted[a.Name] || r.registry.Status(a.Name) != StatusOK {
+			continue
+		}
+		if _, err := r.registry.Run(ctx, a.Name, Params{Since: time.Now().Add(-30 * 24 * time.Hour), Until: time.Now(), Limit: 100}, RunOptions{Fresh: true}); err != nil {
```
aakarim: **Deviation (egress rule, decision 25):** `Refresh` runs every `StatusOK` aggregation, and `tree-size` / `largest-docs` are `StatusOK` because `Requires: ["facts"]` is satisfied once `reg.Bind(log, facts)` runs. So every 5 minutes the Refresher does a **full recursive read of the whole VFS** (`facts.Walk("/")` reads every file's bytes to count lines/tokens). The plan's egress rule says the Refresher "never schedules ContentFacts aggregations (no default period — they are `--fresh`-only unless an operator sets `aggregations.refresh.<name>`)", precisely because this is the one path that will cost money on an S3-backed VFS. Skip aggregations that require `facts` here (or make `Requires: facts` imply `Refresh: never`).

Second, the `Params` are hard-coded (30d window, limit 100, no `Extra`), so the materialised row has a different key from anything a CLI/HTTP caller asks for — see the `paramsKey` comment; as written these refreshes can never be served from the store.

internal/analytics/storage.go
```diff
@@ -0,0 +1,447 @@
...
+	mu  sync.Mutex
+}
+
+func OpenAggregationStore(dir string) (AggregationStore, error) {
+	if err := os.MkdirAll(dir, 0o700); err != nil {
+		return nil, err
+	}
+	return &fileAggregationStore{dir: dir}, nil
+}
+func paramsKey(name string, p Params) string {
```
aakarim, line 448: **Bug (AggregationStore is a write-only cache):** `paramsKey` hashes the full `Params` JSON, and both `analyticsParams` (CLI), `queryParams` (HTTP) and `Refresher.Refresh` fill `Since`/`Until` with `time.Now()` at nanosecond precision. Every call therefore gets a unique key: `Get` never hits, and every `Run` (and each Refresher tick × 8 aggregations) writes a **new** `aggregations/<name>-<hash>.json` that nothing ever deletes. `Invalidate` only removes the current key. Over a day at the 5m default that is ~2300 orphan files, plus one per dashboard page load × aggregation.

Suggest keying on the *relative* window (e.g. `Until - Since` rounded to the minute, or the raw `--since 7d` string) plus `Limit` + sorted `Extra`, and having `Put` replace the previous file for the same name/window. The plan's `Materialized.Window` + `ComputedAt` is then enough for the caller to decide freshness.

internal/analytics/pipeline.go
```diff
@@ -0,0 +1,404 @@
...
+func (p *Pipeline) Run(ctx context.Context) {
+	p.once.Do(func() {
+		ctx, p.cancel = context.WithCancel(ctx)
+		checkpointID := p.readCheckpoint().EventID
+		p.processMu.Lock()
+		p.lastEventID = checkpointID
+		p.processMu.Unlock()
+		go func() {
+			defer close(p.done)
+			var retained []Event
```
aakarim: **Scalability (decision 23: "reads the persisted EventLog from a checkpoint"):** startup reads the **entire** retained log into `retained []Event` (every segment, decompressing sealed ones) to locate the checkpoint ID, then re-folds all of it through `consumePersisted`. With the default `retention: keep forever` this is O(all events ever) memory and time on every boot; the plan's checkpoint is a byte offset into a named segment so resume is a seek. Related: `seen sync.Map` (line 54) stores every event ID for the process lifetime and is never pruned, so the live pipeline's memory grows linearly with traffic. Since the hand-off channel and the file tail are the only two sources and the tail already tracks per-segment offsets (`logCursor`), `seen` could be bounded to IDs at or beyond the current tail offset.

internal/analytics/storage.go
```diff
@@ -0,0 +1,447 @@
...
+		cutoff := time.Now().UTC().Add(-l.opts.Retention)
+		for _, seg := range l.Segments() {
+			if seg.Day.Before(cutoff) {
+				_ = os.Remove(seg.Path)
+			}
+		}
+	}
+	return nil
+}
+func (l *fileEventLog) Close() error { l.mu.Lock(); l.closed = true; l.mu.Unlock(); return nil }
```
aakarim: **Minor (Lifecycle step 5 "close files"):** `Close()` only sets a flag. The plan has `EventLog.Close()` fsync and close the active segment. Related: `Append` does `os.Stat(active+".zst")` + `OpenFile` + `Sync` + `Close` per event; with a persistent active-segment handle (opened on first write for the day, rotated on day change) the fsync stays per-event but the stat/open/close go away, and `Close()` then has something real to do. Not a correctness problem today, just the per-event syscall count on the applier-adjacent path.

pkg/openlore/analytics_plugin.go
```diff
@@ -0,0 +1,254 @@
...
+					if claims, err := s.issuer.Verify(bearerToken(r)); err == nil {
+						id, err = s.identityStore.Resolve(r.Context(), claims)
+						ok = err == nil
+					}
+				}
+				if !ok {
+					id = s.identityFromContext(r.Context())
+					ok = id.IdentityName != "" && id.IdentityName != "guest"
+				}
+				if !ok || s.authEnforced && !s.hasCurrentCapability(id, "lore:analytics:view") {
```
aakarim: **Deviation (decision 12):** the plan mounts `/analytics/*` "only when auth is enforced; returns 404 without `lore:analytics:view` so the routes are indistinguishable from absent" and says "no anonymous/guest access even in open posture". Here the routes are always mounted, and when `s.authEnforced` is false the capability check is skipped entirely — any request whose context identity isn't literally `""`/`"guest"` gets the dashboard, `/analytics/facts`, and aggregations that include principals and commands. The CLI side is stricter (`server.go:1149` only calls `SetAnalytics` when `authEnforced`), so the two surfaces disagree. Suggest `if !s.authEnforced { return nil, nil }` from `PrepareHTTPRoutes` (or register nothing), and require the capability unconditionally inside `auth`.

Also note the routes are hard-coded to `/analytics/...`; the plan writes `<prefix>/analytics/*` — fine if the mux the registrar receives is already prefix-scoped, just confirming.

pkg/openlore/server.go
```diff
@@ -1079,9 +1132,35 @@ func (s *Server) buildSessionShell(id Identity) *shell.Shell {
...
+				fields := map[string]any{"command": usage.Command}
+				if usage.Kind == "unknown_syntax" {
+					eventType = "syntax.unknown"
+					fields = map[string]any{"syntax": usage.Syntax}
+				}
+				s.analytics.Record(context.Background(), s.analyticsEvent(id, eventType, fields))
+			}
 		})
 	}
+	if s.analytics != nil && s.authEnforced && id.IdentityName != "" && id.IdentityName != "guest" && id.policySnapshot != nil && s.hasCapabilityForPolicy(*id.policySnapshot, "lore:analytics:view") {
```
aakarim, line 1164: **Deviation (§Document facts, `CmdContext.EmitMetric`):** `ls --stats` / `ls --json` / `stat --json` and the facts row resolve `ContentFacts` via `analyticsFacts(ctx)` → `Analytics()`, and `SetAnalytics` is only installed for identities holding `lore:analytics:view` under enforced auth. So the agent-facing facts surface — the plan's whole reason for extending `ls` ("so that agents can access it more directly") — is invisible to every ordinary reader and to every open-posture server; `ls --json` silently degrades to plain `ls`. Facts are a read of content the caller can already `cat`, so they need no extra permission; only the `analytics` command (aggregations over other principals' activity) does.

Same root cause blocks Phase 2: the plan's `CmdContext.EmitMetric(eventType, fields)` ("no-op when analytics is off") is how `search.query` and `doc.read` get emitted from *every* session. It isn't in this PR, and the only handle commands have is `Analytics()`, which is nil for non-viewers, so Phase 2 would only ever record admins' searches. Suggest two separate seams on `CmdContext`: `Facts() ContentFacts` (always available when analytics is on) + `EmitMetric(...)` (always available), and keep `Analytics()`/the `analytics` command behind the capability.

pkg/shell/cmds/stat.go
```diff
@@ -27,6 +44,12 @@ func CmdStat(ctx CmdContext, args []string, w io.Writer, errW io.Writer, stdin i
 		if !f.FileModTime.IsZero() {
 			fmt.Fprintf(w, "Modify: %s\n", f.FileModTime.Format("2006-01-02 15:04:05"))
 		}
+		if factsProvider != nil {
```
aakarim: **Egress rule (decision 25: "current facts read the VFS only when a user asks"):** plain `stat` (no `--json`) now always calls `factsProvider.Stat`, which for a directory is a full recursive read of the subtree. `stat /` on an S3-backed docset would read every object. Suggest only computing facts on `--json`/`--stats`, or for regular files only, in the default output.

pkg/shell/cmds/ls.go
```diff
@@ -40,7 +50,25 @@ func CmdLs(ctx CmdContext, args []string, w io.Writer, errW io.Writer, stdin io.
...
 			if longFormat {
-				PrintLong(w, f)
+				if statsFlag && factsProvider != nil {
+					facts, factErr := factsProvider.Stat(context.Background(), p)
+					if factErr != nil {
+						fmt.Fprintf(errW, "ls: %s: %s\n", target, factErr)
+						exitCode = 1
+						continue
+					}
+					fmt.Fprintf(w, "%s %8.0f %8.0f %s\n", f.Mode(), facts.Scalars["lines"], facts.Scalars["tokens"], f.Name())
```
aakarim: Nit: for a single-file target `ls -l --stats` prints `mode lines tokens name` (no size), while directory entries a few lines down print `mode size lines tokens name`. Same flag, different column count. Plan asks for `-l` output to stay POSIX-shaped with the extra columns appended; suggest the same row format in both branches.

pkg/openlore/server.go
```diff
@@ -1707,6 +1815,16 @@ func (s *Server) ListenAndServe() error {
 	return s.srv.ListenAndServe()
 }
 
+func (s *Server) metricsExportHandler() http.Handler {
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", s.metrics.Handler())
+	mux.Handle("/metrics/prometheus", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
aakarim: **Deviation (decision 10 / §Observability exports) — please confirm intent:** the plan puts the Prometheus text exposition **at `/metrics` on `metrics_port`** ("`internal/metrics` counters are replaced by the Aggregator"; the ops note says "Phase 1 makes the yml comment 'Prometheus metrics port' true"), with **exemplars carrying `invocation_id`** on `openlore_command_duration_seconds`. Here `/metrics` keeps serving the JSON blob and the exposition lives at `/metrics/prometheus` with no exemplars. I see this was a deliberate response to the Copilot comment about preserving the JSON contract. If we want to keep JSON, content negotiation on `Accept` (Prometheus sends `text/plain;version=0.0.4` / `application/openmetrics-text`) keeps the conventional path for scrapers without breaking curl users; either way the plan text should be updated to match whichever we pick, and exemplars are still owed by Phase 1.

pkg/openlore/server.go
```diff
@@ -1719,15 +1837,36 @@ func (s *Server) Shutdown(ctx context.Context) error {
...
 		}
 	}
-	if s.srv == nil {
-		return nil
+	if s.analytics != nil {
+		shutdownTimeout := s.config.Analytics.ShutdownTimeout
+		if shutdownTimeout <= 0 {
+			shutdownTimeout = 10 * time.Second
+		}
+		analyticsCtx, cancel := context.WithTimeout(context.Background(), shutdownTimeout)
```
aakarim: **Minor (decision 24 / Lifecycle):** the analytics drain gets `context.WithTimeout(context.Background(), shutdown_timeout)`, so it is *not* bounded by the caller's `ctx`. If transports + write log already used most of the 25s `NotifyContext` budget in `main.go`, this can run up to 10s past it (35s total, past fly's 30s `kill_timeout`). The plan says the budget is "reserved **from** `Server.Shutdown`". Suggest deriving from `ctx` and capping at `shutdown_timeout` (i.e. `min(parent deadline, now+10s)`), or reserving the 10s up front by giving the transports `ctx` minus 10s.

pkg/openlore/analytics_plugin.go
```diff
@@ -0,0 +1,254 @@
...
+						action = "update"
+						break
+					}
+				}
+			} else {
+				action = "delete"
+			}
+			fields := map[string]any{"path": vfs.CleanPath(leaf.Target), "docset": docsetFromPath(leaf.Target), "action": action, "writer": string(writer), "commit_id": info.ID, "commit_hash": info.Hash}
+			if leaf.Write != nil {
+				facts := analytics.ComputeScalars(leaf.Target, leaf.Write.Bytes)
```
aakarim: **Minor (decision 23: "scalar metrics are processed *after* the events are persisted"):** `analytics.ComputeScalars` runs every provider (lines, words, tokenizer) on the post-commit applier goroutine just to obtain `content_hash`. `info.Leaves[i].AfterHash` already carries that hash (once the shadowing bug in `writelog.go` is fixed), or `hashContent(leaf.Write.Bytes)` if you'd rather not depend on `Leaves`. Keeps the write path to a hash only, as the plan intends.

aakarim (commented):
Review against the published plan (`/channel/project-openlore/plans/usage-metrics-and-analytics.md`, fifth pass, decisions 1–27). I built and ran the branch (`go build ./... && go test ./...` green on 1.26) and checked each seam the plan names. Line comments below are ordered by severity; each cites the plan decision it deviates from.

**What conforms (no comments needed):** config shape and defaults (`experimental`, `analytics.*`, `OPENLORE_EXPERIMENTAL`); the event envelope and ULIDs; Recorder → separate Pipeline goroutine with hand-off channel; Registry/Aggregation/Materialized/Status contract and every catalogue aggregation registered (`planned()` stubs for Phase 2); `(commit_id, path)` dedupe; `WriterClassifier` (human iff no actor and the principal is a registered identity); `BlobStore` layout `<hh>/<sha256>` with hash verification and temp+rename; `HistoryCursor{Next, Seek, Position}` checkpointed inside the pipeline checkpoint; `SetCommandObserver` at `execCall` so all transports count and `TotalCommands` moved to dispatch; correlation IDs per transport; `signal.NotifyContext` in main; no MCP tools; dashboard on the HTTP server, not `metrics_port`; `klauspost/compress` zstd sealing; `main.go` shutdown budget.

**Blocking:** one real bug (comment 1) — the commit journal never carries `Leaves`, so `doc.scalars` is never derived in a running server. Everything downstream of decision 25/26 (size-over-time, write-ratio, `openlore_doc_write_scalar_total`) is dead code in prod until it is fixed. The unit tests pass because they hand-build `CommitRecord.Leaves`; the plan's applier-level test ("pre-image capture A→B→delete through the write log") is missing and would have caught it.

**Also missing from the plan's Phase 1 test list:** `Service.Close` honours the deadline and is idempotent; `Record` after `Close` is a counted no-op; `pipeline.enabled: false` still persists and reports `paused`; ship/pipeline lag health series.

**Smaller plan gaps not worth a line comment:**
- `analytics status` prints `Health` only; the plan lists history cursor position, blob store objects/bytes, and gate state.
- `write-ratio` ignores the `path`/`depth` params and returns one row; the plan groups by docset/folder.
- `Pipeline.Lag()` and `Shipper.Lag()` return a zero timestamp; `Health` never reports the last checkpoint the plan wants for the deploy smoke test ("checkpoint equal to log length in the next boot's log line").
- Derived events are appended via `derivedSink` straight to the log rather than back through `Record` (§Storage 5). Arguably better (they can't be dropped), just noting the plan text differs.

pkg/openlore/history_blobs.go
```diff
@@ -0,0 +1,300 @@
...
+	return &fileHistoryCursor{file: f, reader: bufio.NewReader(f), position: from}, nil
+}
+func (c *fileHistoryCursor) Next(ctx context.Context) (CommitRecord, bool, error) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if err := ctx.Err(); err != nil {
+		return CommitRecord{}, false, err
+	}
+	line, err := c.reader.ReadBytes('\n')
+	if errors.Is(err, io.EOF) && len(line) == 0 {
```
aakarim, line 254: Robustness nit on the sequential cursor (decision 25 relies on it being lossless): if `ReadBytes('\n')` returns `io.EOF` with a non-empty partial line (a record whose write is still in flight), the partial bytes are consumed and `json.Unmarshal` fails, and on the next call the reader continues from the middle of that line — that commit is lost to the processor and the next one may fail to parse too. `appendCommitRecord` uses a single `Encode`, so a torn read is unlikely today, but it's cheap to treat `EOF && len(line) > 0` as "not ready": `RestorePosition(c.position)` and return `ok=false`.

pkg/openlore/server.go
aakarim: Clarifying scope after discussing with Adil: the Prometheus exposition must stay on the **internal `metrics_port` listener, unauthenticated, aggregate series only** — never on the external HTTP server behind bearer/passkey auth. This PR already does that (`metricsExportHandler` is only handed to `metrics.StartHandlerServer(metrics_port, …)`; `Service.Aggregator()` is not mounted on the HTTP mux), so nothing needs to move. The remaining asks on this thread are just (a) which path on `metrics_port` serves the exposition — `/metrics` per the plan, or keep `/metrics/prometheus` and update the plan — and (b) the `invocation_id` exemplars. The `Accept` negotiation idea was for that same internal listener, not for exposing it externally.

pkg/openlore/writelog.go
aakarim: Fixed in c905c44. The shadowed variable is gone, leaves are filled before commit-state/journal/post-commit, and an applier-level A→B→delete regression test now verifies before/after hashes and emitted scalar deltas.

pkg/openlore/history_blobs.go
aakarim: The latest published plan (revised 2026-09-13, decisions 25–26 and Storage §4) explicitly retains history/commits.jsonl as the canonical change log carrying committed ChangeSet bytes, keeps history/objects for pre-images, and defers history rotation/blob GC to Phase 3. I therefore kept this architecture rather than switching the cursor to events.jsonl.

pkg/openlore/server.go
aakarim: Fixed in c905c44. Every writable server now opens history/objects and journals commits regardless of whether analytics is enabled; only ScalarProcessor registration remains analytics-gated. Added an analytics-disabled server regression test that verifies both the journal and pre-image blob.

pkg/openlore/history_blobs.go
aakarim: Fixed in c905c44. Disabling blob persistence now retains before_hash and before_size, and ScalarProcessor falls back to the recorded sizes for bytes_delta when the blob is unavailable.

internal/analytics/pipeline.go
aakarim: Implemented in c905c44. Shipper now uploads immutable active-segment byte ranges under events/<day>/<start>-<end>.jsonl, uploads sealed segment objects, persists offsets only after successful Put, retries failures, and reports real byte lag. remote:none remains a no-op while sealing still runs.

internal/analytics/pipeline.go
aakarim: Fixed in c905c44. Scheduled refresh skips facts-backed aggregations entirely; an explicitly named facts refresh runs directly against ContentFacts without creating a shared moving-window materialization. Event-backed refresh keys now match equivalent relative windows.

internal/analytics/storage.go
aakarim: Fixed in c905c44. Materialization keys now use the stable relative window duration (rounded to a second), limit, and deterministically encoded Extra values, so moving-window refreshes replace the same file instead of creating an unbounded sequence. Added a regression test.

internal/analytics/pipeline.go
aakarim: Fixed in c905c44. Pipeline checkpoints now persist per-segment offsets plus processor state, restart incrementally from those offsets, and no longer retain the whole event log in memory. In-process dedupe is bounded to 4096 IDs. Legacy ID-only checkpoints remain stream-compatible.

internal/analytics/storage.go
aakarim: Fixed in c905c44. EventLog now keeps one active segment handle, syncs/closes it on rotation/sealing, and Close performs a real sync and close.

pkg/openlore/analytics_plugin.go
aakarim: Fixed in c905c44. PrepareHTTPRoutes registers no analytics routes unless auth is enforced, and mounted routes unconditionally require lore:analytics:view, returning 404 otherwise. Added an open-posture route-absence regression test.

pkg/openlore/server.go
aakarim: Fixed in c905c44. CmdContext now has separate Facts and EmitMetric seams. Session-scoped facts and metric emission are installed for every session when analytics is enabled, while Analytics() and the analytics command remain capability-gated. Added coverage proving facts access does not grant analytics access.

pkg/shell/cmds/stat.go
aakarim: Fixed in c905c44. Plain stat uses only VFS Stat and no longer computes content facts; recursive/content reads happen only for the explicit --json facts request.

pkg/shell/cmds/ls.go
aakarim: Fixed in c905c44. Single-file ls -l --stats now includes the size column and uses the same mode/size/lines/tokens/name shape as directory rows.

pkg/openlore/server.go
aakarim: Fixed in c905c44. Prometheus exposition is now the unauthenticated aggregate-only /metrics endpoint on metrics_port; the legacy JSON snapshot remains available at /metrics.json. Command-duration summaries include invocation_id as an exemplar, never as a series label.

pkg/openlore/server.go
aakarim: Fixed in c905c44. Analytics shutdown now derives its timeout from the caller's context, so it is capped by whichever deadline is earlier.

pkg/openlore/analytics_plugin.go
aakarim: Fixed in c905c44. doc.write reuses LeafRecord.AfterHash and falls back to the write-path SHA-256 helper only when necessary; it no longer runs scalar providers on the post-commit path.

pkg/openlore/history_blobs.go
aakarim: Fixed in c905c44. HistoryCursor restores its prior file/reader position on EOF with a partial record and waits for the terminating newline. Added a regression test that completes the same partial record and reads it successfully.

PR #92, opened 2026-09-14
Activate phase 2 usage analytics

aakarim:
## Summary
- emit correlated search, document-read, and document-hit analytics events
- activate search-gap and least/most-used content aggregations
- add content-hash-aware line usage and dashboard coverage

## Verification
- `go test ./...`
- `go test -race ./internal/analytics ./pkg/openlore ./pkg/shell/...`
- `git diff --check origin/release/v0.7.0...HEAD`


pkg/shell/cmds/find.go
```diff
@@ -50,11 +52,23 @@ func CmdFind(ctx CmdContext, args []string, w io.Writer, errW io.Writer, stdin i
 			}
 		}
 		fmt.Fprintln(w, p)
+		matchedResults++
+		if !info.Dir {
+			matchedFiles++
+			if metricsEnabled(ctx) {
+				if content, readErr := ctx.FS().ReadFile(p); readErr == nil {
```
aakarim: **Deviation (egress rules + Phase 2 table).** This turns a name listing into a full content download: every file `find` matches is `ReadFile`'d solely to hash it for `doc.hit`. The plan's egress rule is that the VFS is read only when the user asked for that content ("user-initiated and scoped to what they asked for"), and `find` did not ask for bytes — on an S3-backed VFS `find / -name '*.md'` would fetch the whole knowledge base.

It also distorts the analytics: `fullLineRange(content)` makes a name match look like a whole-file read, so `least-used-lines` (which treats a unit as lines actually surfaced) and `most-used-files` will count files an agent merely listed. The plan defines `doc.hit` as "a search that surfaced this doc (one per matched file, `unit.lines` = matched lines)" — a name match surfaces no lines.

Suggest dropping the `doc.hit` emission from `find` entirely and keeping only the `search.query` on line 71 (`filled` already captures whether the name pattern found anything). If we ever want name-match hits, they need a hash the VFS already holds (a `Stat`-level hash), not a read.

pkg/shell/cmds/analytics_metrics.go
```diff
@@ -0,0 +1,107 @@
...
+func metricsEnabled(ctx CmdContext) bool {
+	state, ok := ctx.(metricsState)
+	return !ok || state.MetricsEnabled()
+}
+
+func emitDocMetric(ctx CmdContext, eventType, filePath string, content []byte, lines *analytics.LineRange) {
+	if !metricsEnabled(ctx) {
+		return
+	}
+	sum := sha256.Sum256(content)
```
aakarim: **Deviation (Content identity).** The plan says `content_hash` is "the hex SHA-256 the write system already computes … reads only need the hash the VFS already holds", explicitly to avoid a second hash of the same bytes. The session FS already does this: `readTrackingFS.ReadFile` hashes every read and exposes it via `vfs.ReadTracker.LastReadHash(path)` (forwarded by `writableAliasFS` and `readTransformFS`). So after `ctx.FS().ReadFile(p)` the hash is one map lookup away, and this `sha256.Sum256` re-hashes every `cat`/`head`/`tail`/`grep`/`sed` target a second time on the command path.

Suggest:

```go
hash := ""
if tracker, ok := ctx.FS().(vfs.ReadTracker); ok {
    hash, _ = tracker.LastReadHash(filePath)
}
if hash == "" { // FS without read tracking (tests, plain DirFS)
    sum := sha256.Sum256(content)
    hash = hex.EncodeToString(sum[:])
}
```

That keeps one hash per read and stays byte-identical with the write path's hash and the commit journal.

pkg/shell/cmds/grep.go
```diff
@@ -157,7 +166,13 @@ func CmdGrep(ctx CmdContext, args []string, w io.Writer, errW io.Writer, stdin i
...
 		}
 		lines := strings.Split(string(content), "\n")
-		grepLines(lines, filePath, multiFile)
+		lineHits := grepLines(lines, filePath, multiFile)
+		if len(lineHits) == 0 {
+			return
+		}
+		matchedFiles++
+		matchedLines += len(lineHits)
+		emitDocMetric(ctx, "doc.hit", filePath, content, &analytics.LineRange{Start: lineHits[0], End: lineHits[len(lineHits)-1]})
```
aakarim: **Deviation (Phase 2 table: `unit.lines` = matched lines).** Collapsing the hits into a single `{first, last}` span marks every line in between as read. A grep that hits line 3 and line 400 tells `least-used-lines` that lines 4–399 were surfaced, so the one event source the plan relies on for real line resolution ("only ranged reads and grep hits give real line resolution") over-reports coverage.

`ContentUnit` has exactly one `Lines` range by design, so the fix is one `doc.hit` per contiguous run of matched lines (adjacent hits collapse into one range; gaps produce separate events). `lineHits` is already sorted, so it is a small loop here. Same applies to `sed -n` below.

pkg/shell/cmds/sed.go
```diff
@@ -62,22 +65,85 @@ func CmdSed(ctx CmdContext, args []string, w io.Writer, errW io.Writer, stdin io
...
+					if start == 0 {
+						start = i + 1
+					}
+					end = i + 1
+					break
+				}
+			}
+		}
+		if start > 0 {
+			emitDocMetric(ctx, "doc.read", file.path, file.content, &analytics.LineRange{Start: start, End: end})
```
aakarim: Same span issue as `grep`: `sed -n '3p;400p' file` emits a 3–400 range and `least-used-lines` counts the 396 lines in between as read. Emit one `doc.read` per contiguous run of printed lines instead of first→last.

pkg/shell/cmds/sed.go
```diff
@@ -62,22 +65,85 @@ func CmdSed(ctx CmdContext, args []string, w io.Writer, errW io.Writer, stdin io
 			lines := splitLinesForSed(orig)
 			var buf bytes.Buffer
 			applySedCommands(cmds, lines, quiet, &buf)
+			emitDocMetric(ctx, "doc.read", resolved, orig, fullLineRange(orig))
```
aakarim: **Question / possible deviation.** `sed -i` reads the pre-image only to rewrite it, and the write path already records the change (`doc.write` + commit journal + `doc.scalars`). Emitting a `doc.read` here means every in-place edit also bumps the file's `last_read_at` and `reads`, so files an agent edits frequently will look like the most-*read* files and `least-used-*` will never surface them. The plan lists `doc.read` for "content-reading commands" (`cat`, `head`, `tail`, `sed -n`); I would not count an in-place rewrite as a read. If the intent is to track it, it should probably be a different signal than `doc.read`.

pkg/shell/cmds/analytics_metrics.go
```diff
@@ -0,0 +1,107 @@
...
+	unit := map[string]any{}
+	if lines != nil {
+		unit["lines"] = map[string]any{"start": lines.Start, "end": lines.End}
+	}
+	ctx.EmitMetric(context.Background(), eventType, map[string]any{
+		"path":         vfs.CleanPath(filePath),
+		"content_hash": hex.EncodeToString(sum[:]),
+		"unit":         unit,
+		"docset":       docsetForPath(filePath),
```
aakarim: Minor: two sources of truth for `docset`. This naive first-segment derivation (which yields `""` for root paths and ignores mounts) is overwritten in `Server.buildSessionShell` for `doc.read`/`doc.hit` only, via `(&analyticsPlugin{server: s}).docsetForPath`. Any other event type emitted through `emitDocMetric` later (Phase 4 units, for instance) would silently carry the naive value. Suggest dropping `docset` from the command layer and having the server emitter set `docset` for any event that carries a `path` field, so `doc.write`/`doc.scalars`/`doc.read`/`doc.hit` all agree.

internal/analytics/quality_aggregations.go
```diff
@@ -0,0 +1,436 @@
...
+	Principals  int
+}
+
+type DocUsage struct {
+	Path        string
+	ContentHash string
+	LastReadAt  *time.Time
+	Reads       int
+	Hits        int
+	Scalars     map[string]float64
```
aakarim, line 30: Minor: `ColdUnits` is part of the plan's `DocUsage` ("units with zero reads/hits at current hash") but `FileUsage` never fills it, so any caller relying on it gets an always-empty slice. Either populate it (the per-line usage loop in `leastUsedLines` already computes the cold runs) or drop the field until Phase 4 needs it, so the struct does not promise data it does not carry.

internal/analytics/quality_aggregations.go
```diff
@@ -0,0 +1,436 @@
...
+		last  *time.Time
+	}
+	lines := make([]lineUsage, lineCount)
+	err = src.Scan(ctx, EventFilter{From: p.Since, To: p.Until, Types: []string{"doc.read", "doc.hit"}}, func(e Event) error {
+		if vfs.CleanPath(fieldString(e, "path")) != vfs.CleanPath(filePath) || fieldString(e, "content_hash") != current.ContentHash {
+			return nil
+		}
+		start, end, ok := eventLineRange(e)
+		if !ok {
+			start, end = 1, lineCount
```
aakarim, line 365: Note (ties to the `find` comment): any `doc.read`/`doc.hit` without a `unit` is treated as a whole-file read. That is the intended behaviour for `cat`, but it means every hash-bearing event with an empty unit — including the `find` name-match hits as written — warms every line. Once `find` stops emitting `doc.hit`, the remaining callers all set a range, so this default is fine; flagging it so the coupling is explicit.

aakarim (commented):
Reviewed against the analytics plan (`/channel/project-openlore/plans/usage-metrics-and-analytics.md`, Phase 2 table, Content identity, Content-unit convention, Egress rules). Built and ran `go test ./internal/analytics/... ./pkg/shell/... ./pkg/openlore/...` on `5b6e8bd` — all green.

**Conforms to the plan**
- `TopSearchQueries` / `FileUsage` signatures, `QueryCount` / `DocUsage` shapes (plus `Principals`), `LineRange` JSON tags.
- `top-unfilled-queries` reports a distinct principal *count*, not names.
- `least-used-*` take their inventory from `ContentFacts.Walk(StatOnly)`, tokens from the latest `doc.scalars.after.tokens`, `NULL last_read_at` sorts first ascending; `least-used-lines` is keyed by the current `content_hash` so a changed doc starts cold.
- `least-used-*` now `Require: facts` and the merged Refresher skips facts aggregations on the tick, so nothing walks the VFS on a timer.
- `doc.read` is emitted by the commands via `CmdContext.EmitMetric`, not from the VFS read chain; `search.query` from `grep` and `find` with `filled`/`matched_*`.

**Deviations** are in the line comments, ordered by severity: (1) `find` downloads every matched file's bytes to emit `doc.hit` — an egress-rule violation on S3-backed content and it pollutes line-level stats; (2) `content_hash` is re-hashed with SHA-256 on every read instead of reusing the hash `readTrackingFS` already recorded; (3) `grep`/`sed -n` collapse matched lines into one first→last span, which over-reports coverage in `least-used-lines`; (4) `sed -i` counts the rewrite's pre-image read as a `doc.read`; (5) two `docset` derivations; (6) `DocUsage.ColdUnits` is never populated.

pkg/shell/cmds/find.go
aakarim: Fixed in 87b4e31. `find` no longer reads matched file contents or emits `doc.hit`; it emits only `search.query` with match counts and `filled`.

pkg/shell/cmds/analytics_metrics.go
aakarim: Fixed in 87b4e31. Document metrics now reuse `vfs.ReadTracker.LastReadHash(path)` and retain SHA-256 only as a fallback for filesystems without read tracking.

pkg/shell/cmds/grep.go
aakarim: Fixed in 87b4e31. Grep now emits one `doc.hit` per contiguous run, with a regression test for sparse matches on lines 2 and 4.

pkg/shell/cmds/sed.go
aakarim: Fixed in 87b4e31. Quiet sed output now emits one `doc.read` per contiguous selected run; sparse lines are covered by a regression test.

pkg/shell/cmds/sed.go
aakarim: Fixed in 87b4e31. `sed -i` no longer emits `doc.read`; the write telemetry remains authoritative.

pkg/shell/cmds/analytics_metrics.go
aakarim: Fixed in 87b4e31. The command layer no longer derives `docset`; the server centrally enriches every path-bearing semantic event through configured mount-aware mapping.

internal/analytics/quality_aggregations.go
aakarim: Fixed in 87b4e31. `FileUsage` now derives cold contiguous line units from current-hash read/hit coverage and the latest scalar line count, ignoring stale hashes.

internal/analytics/quality_aggregations.go
aakarim: Addressed in 87b4e31 by removing `find` hits. Empty units remain whole-file semantics, while grep and quiet sed always emit explicit contiguous ranges.

internal/analytics/quality_aggregations.go
```diff
@@ -0,0 +1,490 @@
...
+		if start < 1 {
+			start = 1
+		}
+		if end > lineCount {
+			end = lineCount
+		}
+		for i := start; i <= end; i++ {
+			lines[i-1].reads++
+			last := e.Time
+			lines[i-1].last = &last
```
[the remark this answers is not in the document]
aakarim: Fixed in 538ec93. Per-line timestamps now retain the maximum event time regardless of append order, with an out-of-order regression test.

pkg/shell/cmds/analytics_metrics.go
```diff
@@ -0,0 +1,115 @@
...
+	if !metricsEnabled(ctx) {
+		return
+	}
+	contentHash := ""
+	if tracker, ok := ctx.FS().(vfs.ReadTracker); ok {
+		contentHash, _ = tracker.LastReadHash(filePath)
+	}
+	if contentHash == "" {
+		sum := sha256.Sum256(content)
+		contentHash = hex.EncodeToString(sum[:])
```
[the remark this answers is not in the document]
aakarim: Fixed in 538ec93. Added an exact surfaced-content hashing contract for presentation-transform filesystems. Transformed reads now use the transformed-byte hash (matching ContentFacts and line numbering), while ordinary reads continue reusing the durable ReadTracker hash. Multi-range emissions compute this identity once per file.

pkg/shell/cmds/sed.go
```diff
@@ -69,15 +70,72 @@ func CmdSed(ctx CmdContext, args []string, w io.Writer, errW io.Writer, stdin io
...
+		if !quiet {
+			emitDocMetric(ctx, "doc.read", file.path, file.content, fullLineRange(file.content))
+			continue
+		}
+		var selected []int
+		for i := 0; i < file.lineCount; i++ {
+			global := file.lineOffset + i
+			for _, command := range commands {
+				if command.command == 'p' && sedAddressMatch(command, global+1, len(lines), lines[global]) {
+					selected = append(selected, i+1)
```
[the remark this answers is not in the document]
aakarim: Fixed in 538ec93. Sed now records actual `p` decisions in the main mutable pattern-space execution path instead of re-evaluating original input. The substitution-then-regex-print case is covered by a regression test.

aakarim:
Also addressed all seven suppressed findings from the latest review in 538ec93: directory paths are rejected by least-used-lines; aliases canonicalize for event identity and docset attribution; empty head/tail selections emit no reads; sparse ranges hash once per file; grep excludes the trailing-newline sentinel; stdin grep has no filesystem scope; and tail -n 0 emits no read. Full `go test ./...` and targeted race suites pass.

PR #94, opened 2026-09-15
Implement phase 3 analytics scale and general availability

aakarim:
## Summary

- add SQLite materialization storage and an S3-compatible remote event store with remote rebuild support
- harden the analytics dashboard with pagination, saved windows, CSV exports, and storage/pipeline health
- rotate and compress commit history, retain per-path commit metadata, and add fail-safe blob garbage collection
- remove the experimental gate so `analytics.enabled` alone controls the generally available feature

## Verification

- `nix --extra-experimental-features 'nix-command flakes' develop -c go test ./...`
- `nix --extra-experimental-features 'nix-command flakes' develop -c go test -race ./internal/analytics ./pkg/openlore`
- `nix --extra-experimental-features 'nix-command flakes' develop -c go vet ./internal/analytics ./pkg/openlore ./pkg/shell/cmds`
- rendered and inspected desktop and 390px analytics overview/detail views; verified the detail table horizontally scrolls (`clientWidth=350`, `scrollWidth=531`)


internal/analytics/service.go
```diff
@@ -126,18 +163,66 @@ func (s *Service) Refresh(ctx context.Context, names ...string) error {
...
+	}
 	if s.pipeline == nil {
 		return nil
 	}
 	return s.pipeline.Replay(ctx, from)
 }
+
+// RebuildFromRemote recreates disposable metrics and materializations from the
+// shipped event archive after local analytics data is lost.
+func (s *Service) RebuildFromRemote(ctx context.Context) error {
```
aakarim, line 168: **Deviation (Storage architecture §3: "Rebuild = replay EventLog (or RemoteEventStore after local loss) into a fresh AggregationStore").** The rebuild materialises tables from the remote, but the `Registry` stays bound to the local `EventLog` (`reg.Bind(log, facts)` in `New`). The Refresher's next tick calls `registry.Run(…, RunOptions{Fresh: true})`, recomputes every aggregation from the empty local log under the same `paramsKey`, and overwrites the rebuilt tables. Any dashboard request with `fresh=true`, a different window, or a different `limit` also computes against the empty log, and the Pipeline checkpoint/lag still describe a log with nothing in it.

I reproduced it: after `RebuildFromRemote` then one `service.Refresh(ctx)` (what the 5m tick does), `top-commands` returns `status=ok rows=0` where the existing test expects one `stat` row.

Two ways to make the rebuild stick, either consistent with the plan:
1. **Restore the log, not the tables.** Download the remote objects back into `<dir>/` as local segments (sealed `.jsonl.zst` as-is; concatenate a day's deltas into that day's `.jsonl` when no sealed object exists), then run the ordinary `pipeline.Replay` from a zero checkpoint. Everything downstream (Refresher, `--fresh`, `/metrics`, shipper checkpoint, health lag) is then correct with no special cases, and "after local loss" means what an operator expects.
2. If materialising directly from the remote is kept, `Bind` the Registry to a source that consults the remote when the local log is empty, so the Refresher does not undo the rebuild.

Option 1 is smaller and removes the second code path (`aggregation.Compute` + `store.Put` here duplicates `Registry.run`).

aakarim (commented):
Reviewed against the analytics plan (`/channel/project-openlore/plans/usage-metrics-and-analytics.md`: Phase 3 section, Storage architecture, Rules, Interfaces). Built and ran `go build ./... && go test ./internal/analytics/... ./internal/config/... ./pkg/openlore/ ./pkg/shell/cmds/` on `f4bbf9f` — all green.

**Conforms to the plan**
- `aggregations.store: sqlite` driver behind the frozen `AggregationStore` interface; file store stays the default; unknown driver names are rejected.
- `ship.remote: s3` driver behind the frozen `RemoteEventStore` interface using the plan's object-key scheme (zero-padded byte-range deltas, sealed `.jsonl.zst`); `RemoteSource` adapter reads both forms.
- Gate removal is exactly as specified: the `experimental` config key still parses, only the `ExperimentalEnabled("analytics")` check is gone, banner text removed, `analytics.enabled` alone governs.
- Health now carries last refresh, history cursor position/lag and blob store size; dashboard has CSV download of any `Table`, fixed windows, pagination, and required-parameter handling.
- History rotation writes daily `commits-<day>.jsonl.zst` segments and the `HistoryCursor` recovers across rotation by `LastID`; `history gc` runs under `writeLog.Do` so it is serialised with commits and refuses to delete on a torn journal.
- Per-path commit index is provided by adding `commit_id` to the existing per-file `HistoryRecord` shards and retaining remove records — different mechanism from the plan's SQLite index, but it gives `history <path>` the head pointer the plan asked for.

**Deviations** in the line comments, by severity: (1) `RebuildFromRemote` results are overwritten by the next Refresher tick because the Registry stays bound to the (empty) local log — reproduced; (2) journal rotation + zstd compression runs synchronously on the applier goroutine, i.e. on the commit path; (3) history retention prunes segments regardless of the `HistoryCursor` checkpoint, so unprocessed change scalars can be lost silently; (4) `HistoryCursorLagBytes` decompresses every sealed segment on each health render; (5) `RemoteSource.Scan` reads deltas *and* the sealed object for every day and keeps an unbounded `seen` set; (6) pagination is smuggled into `Params.Extra["_offset"]`, making each page a separate materialisation key.

internal/analytics/service.go
```diff
@@ -126,18 +163,66 @@ func (s *Service) Refresh(ctx context.Context, names ...string) error {
 	return s.refresher.Refresh(ctx, names...)
 }
 func (s *Service) Replay(ctx context.Context, from time.Time) error {
+	if len(s.log.Segments()) == 0 && s.remote != nil {
+		return s.RebuildFromRemote(ctx)
```
aakarim: Related: `Replay(ctx, from)` silently changes meaning when the local log is empty — `from` is ignored and the whole remote archive is fetched. That is surprising for `analytics replay --from …` on a fresh install that happens to have `ship.remote: s3` configured. Suggest making remote rebuild an explicit subcommand/flag (`analytics rebuild --from-remote`) rather than an implicit fallback inside `replay`, so the S3 egress is a deliberate operator action (Egress rules).

pkg/openlore/writelog.go
```diff
@@ -143,15 +144,19 @@ func (l *writeLog) run() {
 		}
 		if committed.HasCommitted() {
 			if l.commitPath != "" {
-				if recordErr := appendCommitRecord(l.commitPath, CommitRecord{ID: commitID, Time: time.Now().UTC(), Attribution: e.attribution, ChangeSet: committed.Committed, Hash: committed.Hash, Leaves: leaves}); recordErr != nil {
+				recordedAt := time.Now().UTC()
+				if rotateErr := RotateCommitJournal(context.Background(), l.commitPath, recordedAt, l.historyRetention); rotateErr != nil {
```
aakarim: **Deviation (Rules: nothing derived runs on the write path; Phase 3: rotation uses "the same `LogOptions` shape as the EventLog").** `RotateCommitJournal` runs on the applier goroutine before every commit record: on every commit it does `os.Stat`, opens the journal to decode the first line, and `ReadDir`s the history directory for `pruneHistory`; on the first commit of a new UTC day it zstd-compresses the whole previous day's journal (which carries the full committed bytes of every change set) synchronously, with `commitJournalMu` held. Every writer on the server is blocked behind that compression, and a failure here is logged after a durable write the caller already thinks is complete.

The EventLog seals off the write path (`Seal` runs from the Shipper tick). The journal should do the same: the applier only appends; a background ticker (or the existing analytics `Shipper`/`Refresher` goroutine, under the server context with a bounded `Close`) seals segments older than today and prunes. The cursor already handles concurrent rotation via `recoverRotation`, so nothing else changes.

pkg/openlore/history_blobs.go
```diff
@@ -296,10 +538,260 @@ func (c *fileHistoryCursor) Position() HistoryPosition {
...
+		}
+		dayText := strings.TrimPrefix(n, "commits-")
+		if len(dayText) < len("2006-01-02") {
+			continue
+		}
+		day, err := time.Parse("2006-01-02", dayText[:len("2006-01-02")])
+		if err != nil {
+			continue
+		}
+		if day.Add(24 * time.Hour).Before(now.UTC().Add(-retention)) {
```
aakarim, line 910: **Deviation (Storage architecture §4/§5: change scalars are derived from the history via a checkpointed cursor, and the log is the durability boundary).** Retention deletes sealed history segments purely by age. If the Pipeline has been paused (`pipeline.enabled: false`) or the server was down for longer than `history.retention`, the `HistoryCursor` checkpoint still points into a segment that no longer exists; `recoverRotation` cannot find `LastID`, falls through to `active, 0`, and every commit in the pruned segments loses its `doc.scalars` silently. Then `history gc` (which only reads retained records) deletes their pre-image blobs, so the loss is permanent.

Retention should never prune past the cursor checkpoint: read the checkpoint (or take a `minRetainedSegment` callback) and keep everything from that segment onward, and surface the backlog as a health value rather than deleting it. Same rule the Shipper follows for the EventLog ("backlog is bounded by local retention" only once shipped).

pkg/openlore/history_blobs.go
```diff
@@ -296,10 +538,260 @@ func (c *fileHistoryCursor) Position() HistoryPosition {
...
+			return 0, err
+		}
+		return st.Size(), nil
+	}
+	zr, err := zstd.NewReader(f)
+	if err != nil {
+		return 0, err
+	}
+	defer zr.Close()
+	return io.Copy(io.Discard, zr)
```
aakarim: `HistoryCursorLagBytes` computes the logical size of every sealed segment after the cursor by fully decompressing it. This runs on every `Service.Health()` call — the dashboard overview, `analytics status`, and the health series — so a year of daily journals (each holding full change-set bytes) is decompressed end-to-end per page view, inside a 2s timeout that will simply truncate the answer. Record the uncompressed size at seal time (a sidecar `.size` file or a `commits-<day>.<bytes>.jsonl.zst` name) and read that; or report lag in segments plus active-file bytes, which is what an operator actually needs.

internal/analytics/storage_remote.go
```diff
@@ -0,0 +1,228 @@
...
+
+func RemoteSource(store RemoteEventStore) EventSource { return NewRemoteSource(store) }
+
+func (s *remoteSource) Scan(ctx context.Context, filter EventFilter, fn func(Event) error) error {
+	keys, err := s.store.List(ctx, s.prefix)
+	if err != nil {
+		return err
+	}
+	sort.Strings(keys)
+	seen := map[string]struct{}{}
```
aakarim: Two rebuild-cost issues (Egress rules):

1. For a day that has a sealed `events-<day>.jsonl.zst`, the deltas are read too and then de-duplicated — roughly 2× the S3 egress of the archive on every rebuild. The plan's key scheme ("the whole sealed object replaces the day's deltas") lets you group keys by day and skip the deltas whenever the sealed object exists.
2. `seen` holds every event ID across the whole archive for the duration of the scan; with per-day grouping it only needs to live for one day (and is unnecessary when the sealed object is present).

pkg/openlore/analytics_plugin.go
```diff
@@ -146,11 +149,14 @@ func queryParams(r *http.Request) analytics.Params {
 	if until, ok := analyticsQueryTime(r.URL.Query().Get("until"), p.Until); ok {
 		p.Until = until
 	}
-	if n, err := strconv.Atoi(r.URL.Query().Get("limit")); err == nil && n > 0 {
+	if n, err := strconv.Atoi(r.URL.Query().Get("limit")); err == nil && n >= 0 && r.URL.Query().Has("limit") {
 		p.Limit = n
 	}
+	if page, err := strconv.Atoi(r.URL.Query().Get("page")); err == nil && page > 1 {
+		p.Extra["_offset"] = strconv.Itoa((page - 1) * p.Limit)
```
aakarim: **Minor deviation (Storage architecture §3: materialisations are keyed by `(aggregation, params)`).** Putting `_offset` into `Params.Extra` makes paging part of the cache key and of the aggregation contract: page 2 is a separate `Materialized` that is computed on first view, cached under its own key, refreshed by nobody (the Refresher only refreshes the default params), and every `Compute` has to remember to call `limitRows`. Pagination is a presentation concern: run the aggregation with the requested window and a `Limit` large enough for the table (or `0` for CSV as you already do), then slice `[offset:offset+pageSize]` in `dashboard`/`aggregation`. One materialisation per window, stable `Total`, and `Params` stays the store's key as designed.

pkg/openlore/server.go
```diff
@@ -1250,7 +1257,23 @@ func (s *Server) buildSessionShell(id Identity) *shell.Shell {
...
+		if canAdmin && s.writeLog != nil {
+			historyDir := filepath.Join(s.config.DataDir, "history")
+			if s.config.DataDir == "" {
+				historyDir = filepath.Join(".openlore", "history")
+			}
+			history.gc = func() (HistoryGCStats, error) {
+				var stats HistoryGCStats
+				err := s.writeLog.Do(context.Background(), func() error {
+					var err error
+					stats, err = GarbageCollectHistoryBlobs(context.Background(), historyDir)
```
aakarim: Minor (Rules: "Context in, bounded drain out"). `history gc` runs under `writeLog.Do(context.Background(), …)`, so a GC over a large `objects/` tree blocks every commit on the server with no deadline and cannot be cancelled by the session that started it. Pass the command's context (and let `GarbageCollectHistoryBlobs` honour it — it already checks `ctx.Err()`), or bound it with the analytics `shutdown_timeout`.

go.mod
```diff
@@ -4,23 +4,43 @@ go 1.26.2
 
 require (
 	github.com/auth0/go-jwt-middleware/v3 v3.2.0
+	github.com/aws/aws-sdk-go-v2 v1.47.0
+	github.com/aws/aws-sdk-go-v2/config v1.33.5
+	github.com/aws/aws-sdk-go-v2/service/s3 v1.113.1
```
aakarim: Note for the record: the plan flagged SQLite as "the first database dependency"; this PR adds the AWS SDK v2 (s3 + config + ~15 indirect modules) alongside `modernc.org/sqlite`. Both are within the Phase 3 scope, just worth calling out in the PR description since the SDK's credential chain (env, shared config, IMDS, SSO) now runs at startup whenever `ship.remote: s3` is set.

aakarim (commented):
Follow-up on scope, after checking with the plan owner.

My earlier review listed the S3 driver as conforming. That was wrong: the plan says in its storage architecture ("RemoteEventStore — interface only in this plan; S3 is the intended first driver") and its explicit non-goals ("No S3 driver in this plan; the `RemoteEventStore` interface and shipper are built and exercised with `remote: none`") that S3 is deferred. The Phase 3 bullet that says otherwise was a drafting error in the plan and is being corrected.

Please pull the S3 driver out of this PR:
- `internal/analytics/storage_remote.go`: drop `S3Config`, `s3API`, `S3EventStore`, `OpenS3EventStore`, `newS3EventStore`; keep `RemoteSource`/`NewRemoteSource` and `eventMatches`.
- `internal/analytics/service.go`: the `case "s3"` branch in `New`; `ship.remote` accepts only `none` (or a `Deps.Remote` injected via `WithAnalyticsRemote`) until an S3 phase.
- `internal/config/config.go` + `openlore.yml.example`: the `AnalyticsS3Config` / `ship.s3` block.
- `go.mod`/`go.sum`: the `github.com/aws/aws-sdk-go-v2/*` and `smithy-go` requires, and the `s3API`-based tests in `storage_drivers_test.go`.

Keep the rebuild path (`RebuildFromRemote`, `RemoteSource`) and its `memoryRemote` tests — that is the "imagine a separate events store" work the plan does want — subject to the Refresher-overwrite issue in my earlier comment.

internal/analytics/service.go
aakarim: Fixed in af16c12: remote rebuild now restores missing events into the canonical local EventLog and invokes the ordinary pipeline replay. The regression test also runs Refresh afterwards and verifies the row, Prometheus metric, and restored local event persist.

internal/analytics/service.go
aakarim: Fixed in af16c12: replay no longer fetches remote data implicitly. Remote recovery is now the explicit `analytics rebuild --from-remote` operator action.

pkg/openlore/writelog.go
aakarim: Fixed in af16c12: the applier only appends now. A server-lifecycle maintenance worker rotates daily; it atomically detaches the active journal under the commit lock, then compresses and prunes outside that lock, with cancellation and bounded shutdown.

pkg/openlore/history_blobs.go
aakarim: Fixed in af16c12: maintenance protects the current HistoryCursor segment and every newer segment from retention. With no active cursor, pruning remains conservative. Regression coverage verifies only segments older than the checkpoint are removed.

pkg/openlore/history_blobs.go
aakarim: Fixed in af16c12: newly compressed history segments publish their logical byte length in a `.size` sidecar. Health reads that metadata; legacy segments use physical size rather than decompression. Cursor checkpoints are also normalized across the plain-to-zstd transition.

internal/analytics/storage_remote.go
aakarim: Fixed in af16c12: RemoteSource groups objects by day, reads sealed zstd objects instead of that day’s deltas, and scopes deduplication state to one day. The test tracks Get calls and verifies both sealed precedence and day-scoped dedupe.

pkg/openlore/analytics_plugin.go
aakarim: Fixed in af16c12: page is no longer added to Params.Extra or the materialization key. The HTTP layer runs/caches the complete requested table and slices it for presentation; CSV remains unpaginated. Page-2 behavior has regression coverage.

pkg/openlore/server.go
aakarim: Fixed in af16c12: history GC receives a context bounded by analytics.shutdown_timeout, and that context is used both while waiting on writeLog.Do and during the object walk. A canceled-GC regression test verifies no blob is deleted.

go.mod
aakarim: Updated in af16c12 per the follow-up scope correction: the S3 driver, S3 config/example, AWS SDK tests, and all AWS/Smithy Go modules have been removed. RemoteEventStore, RemoteSource, the shipper, rebuild flow, and in-memory remote tests remain.

pkg/openlore/analytics_plugin.go
```diff
@@ -222,34 +256,177 @@ func (p *analyticsPlugin) runAggregation(r *http.Request, name string) (analytic
...
+	if page-1 > len(m.Table.Rows)/params.Limit {
+		m.Table.Rows = nil
+		m.Window = params
+		return m, nil
+	}
+	offset := (page - 1) * params.Limit
+	if offset >= len(m.Table.Rows) {
+		m.Table.Rows = nil
+	} else {
+		m.Table.Rows = m.Table.Rows[offset:min(offset+params.Limit, len(m.Table.Rows))]
```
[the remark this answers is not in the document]
aakarim: Resolved in 4da845b. Pagination bounds are checked against the available row count before multiplication, and pagination rendering now compares page numbers rather than multiplying page*limit. A max-int page regression test verifies an empty 200 response instead of a panic.

pkg/openlore/history_blobs.go
```diff
@@ -224,43 +239,196 @@ type historyCursorRestorer interface {
...
 	}
 	if errors.Is(err, io.EOF) {
 		// An append-only journal may be observed between the record write and
 		// its terminating newline. Do not consume that record until complete.
-		if _, seekErr := c.file.Seek(c.position.Offset, io.SeekStart); seekErr != nil {
+		if seekErr := c.openCurrent(); seekErr != nil {
 			return CommitRecord{}, false, seekErr
 		}
-		c.reader.Reset(c.file)
 		return CommitRecord{}, false, nil
```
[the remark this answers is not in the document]
aakarim: Resolved in 4da845b. Unterminated records remain retryable only in the active commits.jsonl; sealed segments now return io.ErrUnexpectedEOF. GC regression coverage verifies sealed corruption aborts before deleting an unreferenced blob.

pkg/openlore/history_blobs.go
```diff
@@ -296,10 +561,381 @@ func (c *fileHistoryCursor) Position() HistoryPosition {
...
+	}
+	if protected.Segment == "" {
+		return false
+	}
+	if protected.Segment == "commits.jsonl" {
+		return true
+	}
+	day, sequence := historySegmentOrder(name)
+	protectedDay, protectedSequence := historySegmentOrder(protected.Segment)
+	return day < protectedDay || day == protectedDay && sequence < protectedSequence
```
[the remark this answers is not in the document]
aakarim: Resolved in 4da845b. When maintenance detaches the active journal, an active protected checkpoint is remapped to the newly compressed segment before pruning. A regression test places pending records after the checkpoint and verifies retention preserves the detached segment.

internal/analytics/service.go
```diff
@@ -54,11 +64,31 @@ func New(cfg config.AnalyticsConfig, deps Deps) (*Service, error) {
...
 			return nil, err
 		}
 	}
+	remote := deps.Remote
+	if remote == nil {
+		switch strings.ToLower(strings.TrimSpace(cfg.Ship.Remote)) {
+		case "", "none":
+		default:
+			err = fmt.Errorf("unsupported analytics remote store %q", cfg.Ship.Remote)
+		}
```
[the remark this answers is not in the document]
aakarim: No S3 backend is intentionally exposed in this phase. The owner’s follow-up review explicitly corrected the scope: ship.remote accepts only none unless a RemoteEventStore is injected through dependencies; RemoteSource and rebuild remain interface-level functionality exercised by in-memory tests. The earlier S3 implementation and configuration were removed in af16c12 for that reason.

internal/analytics/service.go
```diff
@@ -131,13 +162,50 @@ func (s *Service) Replay(ctx context.Context, from time.Time) error {
...
+		existing[event.ID] = struct{}{}
+		return s.log.Append(ctx, event)
+	}); err != nil {
+		return err
+	}
+	if s.pipeline != nil {
+		return s.pipeline.Replay(ctx, time.Time{})
+	}
+	return nil
```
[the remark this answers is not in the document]
aakarim: Resolved in 4da845b. A successful remote replay now invalidates every aggregation materialization, including when the pipeline is disabled. The rebuild test seeds a stale cached result and verifies the first normal Registry.Run returns rebuilt remote data before any refresh.

pkg/openlore/server.go
```diff
@@ -1840,9 +1870,48 @@ func (s *Server) ListenAndServe() error {
...
+	ctx, cancel := context.WithCancel(context.Background())
+	s.historyCancel = cancel
+	s.historyDone = make(chan struct{})
+	go func() {
+		defer close(s.historyDone)
+		maintain := func() {
+			protected := []HistoryPosition{{}}
+			if s.historyPosition != nil {
+				protected[0] = s.historyPosition()
+			}
```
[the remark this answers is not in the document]
aakarim: Resolved in 4da845b. Maintenance now passes no protected checkpoint when analytics has no history cursor, so normal retention remains effective on analytics-disabled writable servers. Pipeline-paused analytics still protects its intentionally stationary cursor, preserving the owner-requested no-data-loss behavior.

aakarim:
Also addressed the valid findings listed in the review summary that did not produce threads: required-parameter overview cards no longer link to invalid CSV exports; legacy active checkpoints recover after rotation; history segment lookup no longer has Scanner’s 64 KiB record cap; GC closes its cursor; and legacy active positions with progress remain prunable. The sealed-plus-late warning does not apply to this shipper: ShipNow seals late old-day files before upload and RemoteSource reads every zstd object for the day. Pipeline-disabled cursor protection remains intentional per the earlier owner review, so paused processing cannot lose history to retention.

PR #95, opened 2026-09-15
feat: add phase 4 plugin analytics extensions

aakarim:
## Summary
- add plugin provider interfaces for analytics emitters, subscribers, processors, scalar providers, tokenizers, and aggregations
- enforce plugin event and aggregation namespaces while exposing public analytics aliases for external Go plugins
- apply plugin scalar providers and exact tokenizers to live facts, history-derived scalars, replay, and plugin aggregations
- document the analytics extension API and cover the end-to-end plugin pipeline

## Verification
- `go test ./...`
- `go test -race ./internal/analytics ./pkg/openlore`

Implements Phase 4 of `/channel/project-openlore/plans/usage-metrics-and-analytics.md`.


internal/analytics/service.go
```diff
@@ -148,11 +151,61 @@ func (s *Service) AddProcessor(processor Processor) {
 		}
 	}
 }
+func (s *Service) AddConsumer(consumer Consumer) {
+	if s.pipeline != nil && consumer != nil {
+		s.pipeline.opts.Consumers = append(s.pipeline.opts.Consumers, consumer)
```
aakarim: **High — data race / plan: "runs after persistence… can never block or fail the emitting command" (Phase 4) and "Additive, never block" (Rules).**

`AddConsumer` (and `AddProcessor` above) append to `s.pipeline.opts.Consumers` / `Processors` with no lock, while the Pipeline goroutine ranges over the same slices in `handleLocked` (pipeline.go:113, 131) under `processMu`. `Server.RegisterPlugin` is public and nothing stops it being called after `ListenAndServe` has called `s.analytics.Start(ctx)` (server.go:1552). Reproduced with a probe that registers a `MetricsSubscriberProvider` after `Start` while events are flowing:

```
WARNING: DATA RACE
Write at 0x00c0001381f8 by goroutine 17:
  analytics.(*Service).AddConsumer()  service.go:156
  openlore.(*Server).registerAnalyticsPlugin()  analytics_extensions.go:82
Previous read at 0x00c0001381f8 by goroutine 19:
  analytics.(*Pipeline).handleLocked()  pipeline.go:113
```

The existing test only registers before `Start`, so `go test -race ./...` doesn't see it. Two acceptable fixes: (a) move the mutation into `Pipeline.AddConsumer/AddProcessor` methods that take `processMu` (and, for processors added after start, restore their checkpoint state the way `readCheckpoint` does at `Run`), or (b) have `Service.AddConsumer/AddProcessor` return an error once `Start` has run and surface it from `RegisterPlugin`. Either keeps the Rules' "additive, never block" guarantee; (a) is what the plan's "plugins consume from the Pipeline" wording implies.

internal/analytics/service.go
```diff
@@ -89,12 +92,12 @@ func New(cfg config.AnalyticsConfig, deps Deps) (*Service, error) {
...
+	providers := append([]ContentScalarProvider(nil), defaultProviders...)
 	if deps.Tokenizer != nil {
 		providers = []ContentScalarProvider{sizeProvider{}, tokenProvider{deps.Tokenizer}}
 	}
 	facts := NewContentFacts(deps.FS, providers...)
-	s := &Service{cfg: cfg, log: log, store: store, facts: facts}
+	s := &Service{cfg: cfg, fs: deps.FS, log: log, store: store, facts: facts, providers: providers}
 	for _, processor := range deps.Processors {
 		if processor != nil && processor.Name() == "doc-scalars" {
 			s.emitted.Store("doc.scalars", true)
```
aakarim, line 103: **Medium — plan Registry: `Status` is `planned` until the aggregation's `Requires` event types have been emitted, then `ok`.**

The emitted set here is a hard-coded list of built-in types plus `doc.scalars` (set only when the `doc-scalars` processor is present). Nothing records types that flow through `Recorder.Record` or `NamespacedSink`, so a plugin aggregation with `Requires: []string{"plugin.<name>.foo"}` never leaves `planned` even after those events are persisted and materialised. Confirmed with a probe: after emitting 200 `plugin.probe.measured` events and closing the service, `Registry().Status("plugin.probe.needs-event")` still returns `planned`. The PR's test avoids this because `plugin.quality.scores` has no `Requires`.

Since Phase 4 is the phase that lets plugins define both the events and the aggregations over them, `Requires` needs to work for plugin types. Simplest fix in the current design: have the Recorder (or the Pipeline, after persistence) record each observed `Type` in `s.emitted`, and persist that set in the pipeline checkpoint so it survives restart — then the hard-coded list can go away too.

internal/analytics/content.go
```diff
@@ -74,8 +75,8 @@ func computeScalars(p string, b []byte, providers []ContentScalarProvider) DocSc
 		for k, v := range provider.Scalars(p, b) {
 			d.Scalars[k] = v
```
aakarim, line 79: **Low — content-scalar contract: keys are flat (`bytes`, `lines`, `words`, `tokens`, …) and the tokenizer label must identify what produced `tokens`.**

`computeScalars` merges providers in registration order with last-write-wins, so a plugin `ContentScalarProvider` that returns a built-in key (`tokens`, `bytes`, `lines`) silently overrides the built-in value while `d.Tokenizer` keeps the previous label (`approx`) — `doc.scalars.tokenizer` would then lie about where `tokens` came from. The `TokenizerProvider` path handles this correctly (it *replaces* `tokenProvider` in `SetTokenizer`); this path doesn't.

Suggest either rejecting plugin providers whose keys collide with built-ins at `AddContentScalarProvider` (return an error from `RegisterPlugin`), or documenting in `docs/plugins.md` that plugins must not emit built-in keys and must use `TokenizerProvider` to change `tokens`. Not blocking; just worth settling before external plugins depend on the behaviour.

aakarim (commented):
Reviewed against the analytics plan (`/channel/project-openlore/plans/usage-metrics-and-analytics.md`, Phase 4 + Rules + Registry sections). Built and ran `go build ./... && go test ./internal/analytics/... ./pkg/openlore/ ./pkg/shell/cmds/` on `40cc88f` — all green. Two of the findings below were confirmed with a throwaway probe test (`-race` report, and `Registry.Status` output).

**Conforms to the plan**
- All six Phase 4 provider interfaces exist with the plan's names and signatures (`MetricsEmitterProvider.SetAnalyticsSink`, `MetricsSubscriberProvider.AnalyticsConsumers`, `MetricsProcessorProvider.AnalyticsProcessors`, `ContentScalarProviderProvider.ContentScalarProviders`, `TokenizerProvider.Tokenizer`, `AggregationProvider` reused), detected by type assertion in `registerPlugin` like the existing middleware providers.
- `NamespacedSink` forces `plugin.<name>.*` on every emitted type, and is a silent no-op when analytics is disabled (base sink nil), matching the "no-op when off" rule.
- Plugin consumers/processors run inside the Pipeline, i.e. after persistence and off the write path; plugin aggregations are prefixed `plugin.<name>.` and appear in list/HTTP/dashboard through the shared Registry/Refresher.
- Scalar providers are used by both `ContentFacts` and the `ScalarProcessor` (`Service.ComputeScalars`), so runtime stat and history-derived scalars stay on the same pure contract; `doc.scalars.tokenizer` is now labelled from the provider.
- `analytics replay` really does recompute past `doc.scalars` with a swapped tokenizer: `ScalarProcessor.ResetForReplay` clears `processed`/`commits` and `ComputeScalars` reads the live provider list.
- Public type aliases in `analytics_extensions.go` are additive and keep external Go plugins off `internal/`.

**Deviations / gaps** are in the line comments, ordered by severity: (1) unsynchronised consumer/processor registration after `Start` (data race, reproduced), (2) plugin aggregations that `Requires` plugin event types are stuck at `planned` forever (reproduced), (3) plugin scalar keys silently override built-ins with a stale tokenizer label, (4) plugin commands emitting through `Shell.EmitMetric` bypass the namespace.

pkg/openlore/analytics_extensions.go
```diff
@@ -0,0 +1,120 @@
...
+	if !analyticsPluginName.MatchString(name) {
+		return fmt.Errorf("analytics plugin %T has invalid name %q", p, name)
+	}
+
+	var base analytics.Sink
+	if s.analytics != nil {
+		base = s.analytics.Sink()
+	}
+	if provider, ok := p.(MetricsEmitterProvider); ok {
+		provider.SetAnalyticsSink(analytics.NamespacedSink(base, name))
```
aakarim, line 111: **Low — gap vs. plan: "every Type is forced to `plugin.<name>.*`" for plugin emitters.**

The namespace is only enforced on the sink handed over here. A plugin that also contributes shell commands can still call `Shell.EmitMetric(ctx, "command.exec", …)` (pkg/shell/shell.go:203) from its command handler and write arbitrary built-in-looking types into the log, which would pollute built-in aggregations like `top-commands`. Fine to leave for a follow-up, but it should be a documented restriction in `docs/plugins.md` now ("plugin code must emit only through the sink from `SetAnalyticsSink`"), or `CmdContext`'s emitter should be namespaced when the executing command was registered by a plugin.

pkg/openlore/analytics_extensions.go
```diff
@@ -0,0 +1,120 @@
...
+	if provider, ok := p.(TokenizerProvider); ok {
+		s.analytics.SetTokenizer(provider.Tokenizer())
+	}
+	if provider, ok := p.(AggregationProvider); ok {
+		prefix := "plugin." + name + "."
+		for _, aggregation := range provider.Aggregations() {
+			if !strings.HasPrefix(aggregation.Name, prefix) {
+				aggregation.Name = prefix + aggregation.Name
+			}
+			if err := s.analytics.RegisterAggregation(aggregation); err != nil {
```
aakarim: Nit: `RegisterAggregation` returns an error on a duplicate name, so two plugins are correctly prevented from colliding via the `plugin.<name>.` prefix — but a *second* `RegisterPlugin` of the same plugin (e.g. after a reload) will fail here after its consumers/processors above have already been appended, leaving the pipeline with duplicate consumers. Consider registering aggregations first (they're the only step that can fail) or making the whole registration all-or-nothing.

internal/analytics/service.go
aakarim: Fixed in 6ffef65. Pipeline.AddConsumer/AddProcessor now synchronize with processing via processMu, and late processors restore their checkpoint state before being appended. Added a post-Start registration test that runs concurrent recording; the targeted race suite is clean.

internal/analytics/service.go
aakarim: Fixed in 6ffef65. The Pipeline now tracks every persisted/derived event type, stores the sorted set in pipeline.checkpoint, restores it on startup, and exposes it to Registry.Status. The plugin test now verifies planned -> ok and that ok survives a restart.

pkg/openlore/analytics_extensions.go
aakarim: Fixed in 6ffef65. Aggregations are now validated and registered atomically before any consumer/processor/scalar contribution is appended, and analytics plugin names are reserved so a second registration fails before side effects. Added duplicate-registration coverage.

pkg/openlore/analytics_extensions.go
aakarim: Documented in 6ffef65: plugin events, including those produced by plugin commands, must use the sink supplied by SetAnalyticsSink. Shell.EmitMetric is explicitly reserved for core commands because it does not apply a plugin namespace.

internal/analytics/content.go
aakarim: Fixed in 6ffef65. bytes, lines, words, and tokens are now reserved to the built-in providers; values for those keys from extension scalar providers are ignored, while custom keys remain additive. Token replacement remains exclusively through TokenizerProvider, and this is documented and tested.

internal/analytics/content.go
```diff
@@ -72,15 +73,33 @@ func computeScalars(p string, b []byte, providers []ContentScalarProvider) DocSc
...
 		for k, v := range provider.Scalars(p, b) {
+			if reservedScalar(k) && !builtinScalarProvider(provider, k) {
+				continue
+			}
 			d.Scalars[k] = v
 		}
-		if provider.Name() == "tokens" {
-			d.Tokenizer = "approx"
+		if provider, ok := provider.(interface{ tokenizerName() string }); ok {
+			d.Tokenizer = provider.tokenizerName()
```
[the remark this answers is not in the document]
aakarim: Fixed in 1a8a640. Directory ContentFacts now derives its tokenizer label from the active provider set, matching the tokenizer used for descendant token totals. Added exact-tokenizer directory coverage.

pkg/openlore/server.go
```diff
@@ -897,6 +900,9 @@ func (s *Server) canonicalChangeSet(cs vfs.ChangeSet) vfs.ChangeSet {
 // per-session at buildSessionShell time, but the post-commit chain is composed
 // once when newWriteLog is constructed.
 func (s *Server) registerPlugin(p any) error {
+	if err := s.registerAnalyticsPlugin(p); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
aakarim: Fixed in 1a8a640. Rejecting metadata-filter validation now runs before analytics wiring, so failed registration installs no sink, processor, provider, consumer, or plugin-name reservation. Added a collision/retry test proving only the accepted consumer receives events.

pkg/openlore/analytics_scalars.go
```diff
@@ -121,6 +126,7 @@ func (p *ScalarProcessor) processRecord(ctx context.Context, record CommitRecord
 				r.Close()
 				facts := p.computeScalars(leaf.Target, b)
 				before = facts.Scalars
+				tokenizer = facts.Tokenizer
```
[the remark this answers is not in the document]
aakarim: Fixed in 1a8a640. ScalarProcessor now replaces the approx fallback only when the computed tokenizer label is non-empty. Added coverage for the custom-provider-only constructor path.

pkg/openlore/analytics_extensions.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	if provider, ok := p.(MetricsSubscriberProvider); ok {
+		for _, consumer := range provider.AnalyticsConsumers() {
+			s.analytics.AddConsumer(consumer)
+		}
+	}
+	if provider, ok := p.(MetricsProcessorProvider); ok {
+		for _, processor := range provider.AnalyticsProcessors() {
+			s.analytics.AddProcessor(processor)
+		}
```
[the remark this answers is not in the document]
aakarim: Fixed in 1a8a640. Plugin processors are wrapped with NamespacedProcessor, which forces Process and Drain outputs plus checkpoint names into plugin.<name>.* while preserving checkpoint/replay/drain capabilities. The end-to-end test now returns an unqualified derived type and verifies its persisted namespace.

aakarim:
Also addressed the three suppressed findings in the latest review in 1a8a640: authenticated/scoped HTTP facts now use Service.NewContentFacts, empty aggregation names are rejected before namespace prefixing, and the corresponding scoped-facts/validation paths have regression coverage.

PR #96, opened 2026-09-16
Fix missing docset roots in lore meta

aakarim:
## Linear

Fixes [OPE-19](https://linear.app/oiya/issue/OPE-19/lore-meta-reports-open-agentheadlong-file-does-not-exist-on-every-run).

Implementation thread: [Amp](https://ampcode.com/threads/T-01a0a8cd-3c81-73f7-993d-23769a791168)

## Diagnosis

`lore meta --filter` stats every configured filter root and treated an absent docset root like any other filesystem failure. In the current code this aborts the command with exit 1, preventing records from later valid roots from being returned. `lore docsets` also gave no indication that the configured root was missing.

## Fix

- Skip roots only when `Stat` reports `fs.ErrNotExist`; preserve failures for all other filesystem errors.
- Add an `absent` attribute to missing roots in `lore docsets`, keeping the configuration problem discoverable without contaminating meta NDJSON output.
- Cover missing-root continuation, non-not-found error propagation, and docset output.

## Verification

- `go test ./pkg/shell/cmds`
- `go test -race ./...`


aakarim:
Reviewed the three suppressed Copilot findings; all represented real gaps and are addressed in commit 2aa8c42:

- `vfs.ErrNotFound` now participates in `errors.Is(err, fs.ErrNotExist)` while preserving its existing message.
- `MergeFS.resolve` returns that typed VFS not-found error instead of a plain string error.
- `lore docsets` now returns a visible failure for non-not-found `Stat` errors instead of reporting the root as healthy.
- Regression tests exercise VFS-originated missing errors and operational `Stat` failures.

Verification: `go test ./pkg/vfs ./pkg/openlore ./pkg/shell/cmds` and `go test -race ./...` both pass.

PR #102, opened 2026-09-17
Fix PWA link and mobile tree navigation

aakarim:
## Summary
- keep rendered Markdown links in the current PWA page instead of opening a new browser tab
- keep the mobile folder sheet open while expanding and collapsing tree branches
- add regression coverage for both navigation behaviors

## Verification
- `npm test` (30 tests passed)
- `npm run check`
- `npm run build`

Fixes OPE-54


aakarim:
Addressed the review note in ecb6043: protocol-relative external URLs now use the same classification as other web links, remove any inherited `target="_blank"`, and have dedicated regression coverage. Reverified with `npm test` (30 passed), `npm run check`, and `npm run build`.

